PR #23, opened 2014-12-15
Escape password in URL to support special chars.

mven:
Special chars, such as #, in a Cloudant password were resulting in a generic authentication error when trying to connect to Cloudant during the configuration process. Escaping the password in config.url variable fixed the issue.


jhs:
Thanks!

Fixed in 2d264af2, issue #32

PR #24, opened 2014-12-17
Fix complete example to use the 'me' variable instead of the string 'me'

mven:
Typo fix so that the Cloudant call is using the variable instead of the string.


jhs:
Oh, thanks! Good catch!

PR #30, opened 2015-02-08
Added support for _security Cloudant endpoints

### LINKED ISSUE #6: Use wilson api's not to be deprecated old ones
https://github.com/cloudant/nodejs-cloudant/blob/master/cloudant.js#L59 and https://github.com/cloudant/nodejs-cloudant/blob/master/cloudant.js#L64 should call the relevant wilson API's (which live inside the users account) rather than the old api server (which will be deprecated one day...). @evansde77 or @ksnavely can point to docs. These are now in use in the dashboard, so can be considered "in production".

glynnbird:
Fixes #6 

The current release uses the deprecated APIs for assigning permissions to users. I've kept these calls the same, but added a deprecation warning and added two new function calls:

* cloudant.view_security(db, callback)
* cloudant.set_security(db, security, callback)

I've followed the same verbs as the other API calls and haven't attempted to hide what the API is doing e.g. I haven't tried to abstract e.g. cloudant.set_role(db, user, roles), for instance.

The functions are wangled into Nano using the same fix_request as the other Cloudant-specific functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,14 +233,18 @@ Output:
 Next, set access roles for this API key:
 
 ~~~ js
-  // Set read-only access for this key.
-  var db = "my_database"
-  cloudant.set_permissions({database:db, username:api.key, roles:['_reader']}, function(er, result) {
+
+  
+  // Set the security for three users.
+  var db = "my_database",
+    roles = [ '_reader', '_writer' ];
+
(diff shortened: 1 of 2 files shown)


jhs:
Ha ha, I just noticed the word "wrangled." Made me laugh. You are seeing the sausage being made.

Issue #32, opened 2015-03-22
Bug if password includes certain symbols?


cmawhorter:
I'm just testing/playing around and thus didn't dive any deeper, but running the first example in the readme that lists databases failed for me:

```
Error: getaddrinfo ENOTFOUND [my username]
    at errnoException (dns.js:44:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:94:26)
```

After confirming everything else, on a whim I simplified my password and it worked.

Didn't bother narrowing but my failing password was 32 characters long and all alpha numeric except for these symbols:  &$#

glynnbird:
Thanks for reporting this @cmawhorter. I traced this to the way that passwords were represented in the URL. Such values need encoding with calls to `encodeURIComponent` to ensure that characters like & # and ? do not break the URL. See PR above.

cmawhorter:
Yay.  That looks like it'll work, but I'd recommend replacing all that concat with [url.format](https://nodejs.org/api/url.html#url_url_format_urlobj), which would probably fix other gremlins beyond this one.

jhs:
Fixed in 2d264af243be8600f264b12c2a1ee82173aa4847

PR #33, opened 2015-03-22
Encoded URL components correctly. Fixes issue #32

glynnbird:
If the password contained an @ or ?, then the URL would not have been formed properly. Fixed by using EncodeURLComponent around such data.


jhs:
Whoops. @glynnbird the string to auto-close is (IIRC) "fixes #32" or "closes #32". Anyway I'll close it manually

PR #40, opened 2015-05-07
Fixing db.insert signature in ToC

nfriedly:
The signature and link were both incorrect for the `db.insert` function in the Table of Contents


jhs:
Thanks! Good catch.

Issue #39, opened 2015-05-06
We need a better way to determine the nano.js version being used (and update nano.js)


conorgil:
I see that nano.js is not listed as a dependency in the `package.json` file and I assume that is because changes were made to the library to support Cloudant functionality. However, this makes it very difficult to determine which version of nano.js is actually being wrapped.

Looking at the history of the included `nano.js` file, I see that the commit at the time it was taken from nano.js was `74ec376c903c22fe39c4560d28da886f6b0a3b8a`. Looking in the nano history, that commit is from May 2014 (over a year ago) and was version `5.9.1` /`5.10.0`. Since then, nano has had many commits and is already up to version `6.1.2`.

Is there a better way to:
1) indicate the version of nano that we wrapped
2) pull in updates from the nano project into this one

Perhaps we could remove the inline `nano.js` file from this projects, fork the `nano.js` repo and introduced the changes required for the Cloudant functionality in that fork, and list that fork as a dependency in the `package.json` file for this project? That might make it easier to merge changes from the `nano.js` project into the fork to get the latest bug fixes, etc? Thoughts?

jhs:
Yes, great points!

In the short term, I am merging in the latest Nano release, and cutting a 1.1 release of this library.

This project is indeed a fork of Nano. However, @glynnbird is right that, really, Nano is our dependency. There are some API changes to get that to work, but that is the goal for v2.0. Then it will be much more clear what is going on.

jhs:
Hi, @conorgil!

Version 1.0.0 is out. Please install it from npm and let us know if you have problems.

For now, 1.x is a *fork* of Nano. However, we are immediately moving into our 2.0 milestone, which intends to do the reasonable thing I should have done since the beginning: make Nano a dependency, and build our features by wrapping it.

So, yes you are completely correct. Having just merged with several months of Nano changes, it was difficult. That is why we are embarking to do this correctly for v2.0. Version 2 may have some API changes, but we do not anticipate many; however it will work with Nano in a more developer-friendly way.

conorgil:
Sounds great!

PR #43, opened 2015-05-12
nodejs-cloudant v2

glynnbird:
I propose we 

* use Nano as a simple dependency (see package.json)
* inject functions into the top of Nano and into the 'db' level
* remove the asynchronous method of starting up the library - this allows folks to use Nano & Cloudant libraries interchangeably 

This tidy up
* only has one dependency - Nano
* reduces this library to 140 lines of code
* makes it a drop in replacement for Nano (which is to be an Apache project now and the official lNode library for CouchDB)

Further work
* tests - the current tests are copied from Nano. Need we duplicate these. Perhaps we should just have the tests for our functions
* documentation - perhaps we should only document the Cloudant-specific functions here, just link to the Nano docs for the main part

Fixes issue #39


bradley-holt:
I haven't done a code review, but +1 conceptually.

glynnbird:
I've rebased my PR so that it doesn't clash with Jason's 1.0.0 work

glynnbird:
Removed some files that weren't required.

PR #75, opened 2015-06-29
Add missing single quotes from README.md

navargas:
Syntax highlight was incorrect due to missing quotation mark.


glynnbird:
Thanks @navargas!

PR #130, opened 2016-05-04
Removes a trailing `/` from a manually specified URL.

### LINKED ISSUE #129: Mixing cfenv with cloudant causes misleading error
## Scenario

1. Using [`cfenv`](https://www.npmjs.com/package/cfenv) (e.g. running on Bluemix)
2. Using a hosted Cloudant instance
3. Get cloudant URL credentials from the cloud foundry environment through: `appEnv.getServiceURL()`
4. Initialize cloudant and run `db.index()`

## Not-working example

```js
// Init
const cfenv = require('cfenv');
const appEnv = cfenv.getAppEnv();

const Cloudant = require('cloudant');
const cloudant = Cloudant(appEnv.getServiceURL('Cloudant Storage'));

// Usage
const db = cloudant.db.use('my-db');
db.index((err, result) => {
  if (!!err) throw err; // THROWS
  console.log("Result: %o", result);
});
```

This throws an error with "Database does not exist." The reason is that `appEnv.getServiceURL()` returns a URL that ends with `/`, due to using the `url` library. 

## Proposed solution

In the `Cloudant` initialization, look for a trailing `/` and remove it. This error propagates into `nano` but it happens within a "typical" use case with Cloudant. 

Work-around is: 

```js
const cloudant = Cloudant(appEnv.getServiceURL('Cloudant Storage').slice(0, -1));
```

Nava2:
Fixes #129.

Tests:
`npm test`
screen shot 2016-05-04 at 12 16 45

I noticed some of the other tests do fail, though.

### DIFF
--- a/lib/reconfigure.js
+++ b/lib/reconfigure.js
@@ -12,6 +12,7 @@ var url = require('url');
 module.exports = function(config) {
   config = JSON.parse(JSON.stringify(config)); //clone
 
+  var outUrl; 
   // if a full URL is passed in
   if (config.url) {
     
@@ -36,7 +37,7 @@ module.exports = function(config) {
       config.url = url.format(parsed);
     }
     
-    return config.url;
+    outUrl = config.url;
   
   } else {
     // An account can be just the username, or the full cloudant URL.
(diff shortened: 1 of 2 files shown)


jhs:
Hi! Thank you very much for the PR. Merged!

PR #131, opened 2016-05-21
ensure that the nano.request object is returned to allow streaming

glynnbird:
I noticed in our cloudant.js code we were doing:

```
function x(callback) {
  nano.request({something:true}, callback);
}
```

That's fine unless you want to stream the output 

```
cloudant.x().pipe(res)
```

So I needed to modify our code to ensure that we always return the nano request object.


jhs:
Thanks. I will test this and then cut a release if/when it passes.

PR #144, opened 2016-07-19
Update account matching regex to allow dashes in account name

### LINKED ISSUE #141: reconfigure.js does not correctly handle Bluemix accounts
In Bluemix, Cloudant service usernames are like "xxxxxxxx-bluemix" and hosts like "xxxxxxxx-bluemix.cloudant.com"

However if I initialize Cloudant with:
```
var cloudant = require('cloudant')({
  account: "1234-4567-bluemix.cloudant.com",
  password: "a0a893028302839382028302830"
});
```

When I try to query a db it fails with error 503. The error shows it tries to access bluemix.cloudant.com:
```
     uri: 'https://XXXXXX:XXXXXX@bluemix.cloudant.com/openwhisk-darkvision/_all_docs' },
```

Basically it did not correctly extract the account name from the hostname I gave which it should be able to do according to the header in reconfigure.js:

```
// reconfigure deals with the various ways the credentials can be passed in 
// and returns an full URL
// e.g. { account:"myaccount", password: "mypassword"}
// or   { key: "mykey", password: "mykey", account:"myaccount"}
// or   { key: "mykey", password: "mykey", account:"myaccount"}
// or   { account:"myaccount.cloudant.com", password: "mykey"}
// or   { account: "myaccount"}
// or   { url: "https://mykey:mypassword@myaccount.cloudant.com"}
```

Seems like the regexp https://github.com/cloudant/nodejs-cloudant/blob/master/lib/reconfigure.js#L46
```
config.account.match(/(\w+)\.cloudant\.com/);
```
does not correctly handle the "-" in the hostname.

The simple workaround is to initialize Cloudant with the username, not the host, even if reconfigure.js was supposed to handle it. This works:
```
var cloudant = require('cloudant')({
  account: "1234-4567-bluemix",
  password: "a0a893028302839382028302830"
});
```
or even simpler, with the "url" given in the Bluemix credentials
var cloudant = require('cloudant')({
  url: "<value-from-vcap-services>"
});

bjustin-ibm:
Resolves #141

### DIFF
--- a/lib/reconfigure.js
+++ b/lib/reconfigure.js
@@ -43,7 +43,7 @@ module.exports = function(config) {
     // An account can be just the username, or the full cloudant URL.
     var match = config.account &&
                 config.account.match &&
-                config.account.match(/(\w+)\.cloudant\.com/);
+                config.account.match(/([^.]+)\.cloudant\.com/);
     if (match)
       config.account = match[1];
 
--- a/tests/reconfigure.js
+++ b/tests/reconfigure.js
@@ -93,5 +93,14 @@ describe('Reconfigure', function() {
     url.should.equal("https://mykey:mypassword@mydomain.cloudant.com");
     done();
   });
-  
+
+  // Issue cloudant/nodejs-cloudant#141
+  it('Allows account names with dashes', function(done) {
+    var credentials = { account: "this-account-has-dashes.cloudant.com", password: "bacon" };
+    var url = reconfigure(credentials);
+    url.should.be.a.String;
+    url.should.equal("https://this-account-has-dashes:bacon@this-account-has-dashes.cloudant.com");
+    done();
+  });
+
 });


glynnbird:
Thanks for the PR @bjustin-ibm!

bjustin-ibm:
@glynnbird Do you have a sense of when this will get published to npm?

glynnbird:
We're hoping to release 1.5.0 in the next couple of weeks, in the meantime, you can always use 'https://github.com/cloudant/nodejs-cloudant' in your package.json?

bjustin-ibm:
Thanks for the info and suggestion. As it turns out we have a workaround (initialize directly with the cloudant url) and we need to carefully control the specific versions of npm packages we are using.  In the meantime, I'll keep an eye out for 1.5.0 ;)

PR #197, opened 2017-07-10
Bump request timeout in tests/api.js

smithsz:
## What
Bump request timeout in `tests/api.js`.

## How
Specify `requestDefaults` when instantiating a request client.

## Testing
Change updates API unit tests.


smithsz:
I'm still getting the odd ECONNRESET error. I think the test infrastructure must be at fault here.

PR #198, opened 2017-07-11
More Jenkins test fixes

### LINKED ISSUE #90: Test suite fails against Cloudant server (minor)
The Gzip test fails when running against the real server. Since AFAIK I am the only person with access to that account, obviously it is not a huge problem since nobody could actually see this besides me anyway.

However, obviously we should fix this minor bug. (I think it's a bug with the test itself, it is not forgiving enough since Cloudant cannot actually gzip as asked.)

But we should also sort out the nodejs.cloudant.com account.

### LINKED ISSUE #179: Retry plugin leaks "Typed Arrays" when used to download attachments
In my app https://github.com/IBM-Bluemix/openwhisk-darkvisionapp I found that if I use the retry plugin and download an attachment from the database, the node.js process memory is increasing up to the point where it gives an out of memory. My node process is given 256MB (it runs inside an OpenWhisk action).

I tried to isolate the issue 

If I use `visionDb.attachment.get('123456', 'video.mp4').pipe(outputStream)`, I can see the memory increasing during the download. visionDb has the retry plugin enabled.

If I use request directly `request.get('http://cloudant/mydb/123456/video.mp4').pipe(outputStream)` everything runs smoothly. Same if I remove the `retry` plugin.

To reproduce the issue, put a 200MB attachment in a Cloudant document, then download the document with only 256MB allocated to node.js `node --max_old_space_size=256  downloadtest.js`

```
const fs = require('fs');
const output = fs.createWriteStream('test.mp4');

const Cloudant = require('cloudant');
const mydb = Cloudant({
  url: 'https://XXX:YYY@host',
  plugin: 'retry',
  retryAttempts: 10,
  retryTimeout: 500
}).db.use('medias');

mydb.attachment.get('yourid', 'video.mp4')
  .pipe(output)
  .on('finish', () => {
     console.log('write complete');
  })
  .on('error', (err) => {
    console.log('error while writing', err);
  });
```

The reference to "Typed Arrays" in the title comes to the observation I made when running the chrome debugger with `node --max_old_space_size=256 --inspect --debug-brk``` and recording heap profile.

smithsz:
## What
- Add/delete 'alice' database for 'Cookie Authentication' test suite.
- Delete 'alice' database after running 'Getting Started' test suite.
- Skip 'Password authentication' tests if NOCK_OFF enabled.
- Update HTTP agent default, namely `keepAliveMsecs: 30000` and `maxSockets: 6`.
- Bump mocha timeout, 20s -> 60s.
- Fix 429 retry plugin & add unit tests.

## Issues
Fixes #90.
Fixes #179 .


plugins/retry.js
```diff
@@ -32,48 +30,63 @@ module.exports = function(options) {
...
-        });
+        var thisRequest = request(req, callback);
+        thisRequest
+          .on('error', function(err) {
+            s.emit('error', err);
+            s.end();
+            done();
+          })
+          .on('response', function(r) {
+            if (r.statusCode === 429 && attempts < maxAttempts) {
```
tomblench: Some of our other libraries have an increasing timeout between 429 failures, do you think this is something we should do to be consistent with them?

ricellis: I think it still does: https://github.com/cloudant/nodejs-cloudant/pull/198/files#diff-0bb0287dc9efec7bdccb2fbde53f5a7bR86

tomblench: thanks @ricellis - didn't see that!

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+      var req = {
+        url: SERVER + '/foo',
+        auth: { username: ME, password: PASSWORD },
+        method: 'GET'
+      };
+      cloudantClient(req)
+        .on('error', function(err) {
+          assert.fail(`Unexpected error: ${err}`);
+        })
+        .on('resonse', function(resp) {
```
tomblench: Typo here so I guess this assert didn't fire

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+      };
+      cloudantClient(req, function(err, resp, data) {
+        assert.equal(err, null);
+        assert.equal(resp.statusCode, 200);
+        assert.ok(data.indexOf('"doc_count":0') > -1);
+      })
+        .on('error', function(err) {
+          assert.fail(`Unexpected error: ${err}`);
+        })
+        .on('resonse', function(resp) {
```
tomblench: also typo here

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+      };
+      cloudantClient(req, function(err, resp, data) {
+        assert.equal(err, null);
+        assert.equal(resp.statusCode, 429);
+        assert.ok(data.indexOf('"error":"too_many_requests"') > -1);
+      })
+        .on('error', function(err) {
+          assert.fail(`Unexpected error: ${err}`);
+        })
+        .on('resonse', function(resp) {
```
tomblench: also typo here

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+    cloudantClient(req, function(err, resp) {
+      assert.equal(err, null);
+      assert.equal(resp.statusCode, 200);
+      mocks.done();
+      done();
+    });
+  });
+
+  describe('with callback only', function() {
+    it('performs request and returns response', function(done) {
```
ricellis: I think this test could use a comment about why it runs even with `NOCK_OFF` when the others don't. Somethin to the effect of, "tests only a single request with a good response so can run against a real server" or whatever.

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+      var req = {
+        url: SERVER + '/foo',
+        auth: { username: ME, password: PASSWORD },
+        method: 'GET'
+      };
+      cloudantClient(req)
+        .on('error', function(err) {
+          assert.fail(`Unexpected error: ${err}`);
+        })
+        .on('resonse', function(resp) {
```
ricellis: Did you get all these typos?

test/plugins/retry429.js
```diff
@@ -0,0 +1,558 @@
...
+        assert.equal(err, null);
+        assert.equal(resp.statusCode, 429);
+        assert.ok(data.indexOf('"error":"too_many_requests"') > -1);
+        mocks.done();
+        done();
+      });
+    });
+  });
+
+  describe('with listener only', function() {
```
ricellis: Might be nice to add a test to this lot that pipes the response stream to ensure that the path through the pass through works as expected.

smithsz: Added in 0dc34f1063b6ca1214fae4493a59788db5acf9e8.

test/plugins/retry429.js
smithsz: Added in 2136bc517a3764eb1181b520603237b102f68456.

PR #200, opened 2017-07-28
Add event pipe

smithsz:
## What
Add `EventPipe` object to forward events from a source emitter to a target emitter.

## How
Store fired events in an array until `.resume()` is called. Then replay captured events (and any new events received) on the target emitter.

## Testing

### DIFF
--- a/test/eventpipe.js
+++ b/test/eventpipe.js
@@ -1,0 +1,89 @@
+// Copyright © 2017 IBM Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


test/eventpipe.js
```diff
@@ -0,0 +1,89 @@
...
+      .on('two', function(x) {
+        receivedEvents.push(x);
+      })
+      .on('three', function(x) {
+        receivedEvents.push(x);
+        assert.deepEqual(receivedEvents, sentEvents);
+        done();
+      });
+
+    ep.resume();
```
ricellis: Worth adding a comment that this is necessary because `EventPipe` starts in `paused` state?

smithsz: good thinking, will do

PR #208, opened 2017-08-29
Add retry plugins

smithsz:

## What
Implement following plugins:
- `retry429` (aka `retry`)
- `retry5xx`
- `retryerror`

## Testing

### DIFF
--- a/plugins/retry429.js
+++ b/plugins/retry429.js
@@ -1,0 +1,37 @@
+// Copyright © 2017 IBM Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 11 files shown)


ricellis:
@smithsz the build is broken for this (and #209 with what looks like the same error)

plugins/retry429.js
```diff
@@ -0,0 +1,44 @@
...
+const BasePlugin = require('./base.js');
+
+/**
+ * Retry 429 response plugin.
+ *
+ * @param {any} client - HTTP client.
+ * @param {any} opts - Plugin options.
+ */
+function Retry429Plugin(client, opts) {
+  Retry429Plugin.super_.apply(this, arguments);
```
ricellis: Any reason this approach is preferred over the ES2105 `class`/`constructor` approach? Personally I find it easier to follow, and it is pretty widely supported now. Although I guess the Node engine `0.8.0` which the `package.json` declares as a minimum is too early for that. I think we only test on the LTS 4, LTS 6 and current now, so maybe we can update that minimum engine anyway.

PR #209, opened 2017-08-29
Add cookieauth plugin

smithsz:

## What
Add `cookieauth` plugin.

## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,9 @@
     "async": "2.1.2",
     "cloudant-nano": "6.5.0",
     "debug": "2.2.0",
-    "request": "^2.81.0"
+    "lockfile": "1.0.3",
+    "request": "^2.81.0",
+    "tmp": "0.0.33"
   },
   "devDependencies": {
     "eslint": "^3.19.0",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "async": "2.1.2",
     "cloudant-nano": "6.5.0",
     "debug": "2.2.0",
-    "lockfile": "1.0.3",
+    "lockfile": "^1.0.3",
     "request": "^2.81.0",
     "tmp": "0.0.33"
   },
--- a/plugins/base.js
+++ b/plugins/base.js
@@ -17,9 +17,10 @@
  * Cloudant base plugin.
  *
  * @param {Object} client - HTTP client.
+ * @param {Object} cfg - Client configuration.
  */
 class BasePlugin {
-  constructor(client) {
+  constructor(client, cfg) {
     this._client = client;
   }
 
--- a/plugins/base.js
+++ b/plugins/base.js
@@ -13,6 +13,12 @@
 // limitations under the License.
 'use strict';
 
+const debug = require('debug')('cloudant:plugins:base');
+const lockFile = require('lockfile');
+const tmp = require('tmp');
+
+function noop() {}
+
 /**
  * Cloudant base plugin.
  *
@@ -22,6 +28,32 @@
 class BasePlugin {
   constructor(client, cfg) {
     this._client = client;
+    this._lockFile = tmp.tmpNameSync({ postfix: '.lock' });
+  }
+
+  // Acquire a file lock on the specified path. Release the file lock on
+  // completion of the callback.
+  withLock(opts, callback) {
+    var self = this;
+    if (typeof opts === 'function') {
+      callback = opts;
+      opts = {};
+    }
+    debug('Acquiring lock...');
+    lockFile.lock(self._lockFile, opts, function(error) {
+      if (error) {
+        callback(error, noop);
+      } else {
+        callback(null, function() {
+          // Close and unlink the file lock.
+          lockFile.unlock(self._lockFile, function(error) {
+            if (error) {
+              debug(`Failed to release lock: ${error}`);
+            }
+          });
+        });
+      }
+    });
   }
 
   get id() {
--- a/plugins/cookieauth.js
+++ b/plugins/cookieauth.js
@@ -14,40 +14,10 @@
 'use strict';
 
 const debug = require('debug')('cloudant:plugins:cookieauth');
-const lockFile = require('lockfile');
 const request = require('request');
-const tmp = require('tmp');
 const u = require('url');
 
 const BasePlugin = require('./base.js');
-
-function noop() {}
-
-// Close and unlink the file lock.
-function releaseLock(lockFilePath) {
-  lockFile.unlock(lockFilePath, function(error) {
-    if (error) {
-      debug(`Failed to release lock: ${error}`);
-    }
-  });
-}
-
-// Acquire a file lock on the specified path. Release the file lock on
-// completion of the callback.
-function withLock(lockFilePath, cfg, callback) {
-  var opts = {
-    stale: cfg.cookieLockStaleMsecs || 1500,  // default: 1.5secs
-    wait: cfg.cookieLockWaitMsecs || 1000     // default: 1sec
-  };
-  debug('Acquiring lock...');
-  lockFile.lock(lockFilePath, opts, function(error) {
-    if (error) {
-      callback(error, noop);
-    } else {
-      callback(null, function() { releaseLock(lockFilePath); });
-    }
-  });
-}
 
 /**
  * Cookie Authentication plugin.
@@ -60,7 +30,6 @@ class CookiePlugin extends BasePlugin {
     self.baseUrl = null;
     self.cookieJar = request.jar();
     self.credentials = {};
-    self.lockFile = tmp.tmpNameSync({ postfix: '.lock' });
     self.useCookieAuth = true;
 
     if (typeof cfg.baseUrl !== 'undefined' &&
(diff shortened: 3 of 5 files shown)


plugins/cookieauth.js
```diff
@@ -13,167 +13,90 @@
...
+  CookiePlugin.super_.apply(this, arguments);
+  this.baseUrl = null;
+  this.cookieJar = request.jar();
+  this.credentials = null;
+  this.refreshRequired = true;
+}
+CookiePlugin.super_ = BasePlugin;
+CookiePlugin.prototype = Object.create(BasePlugin.prototype);
+CookiePlugin.prototype.onRequest = function(state, req, callback) {
+  req.jar = this.cookieJar;
```
ricellis: Can we check for the existence of the cookie and that it isn't expired already so we can preemptively request the `_session` and avoid the `401` round trip if we don't have one or it isn't valid. I think `refreshRequired=true` as the starting point handles the initial request, but if we can check the jar that would be good.

We will need this behaviour more for IAM sessions, but it is beneficial here too, probably it can be extracted out to some shared code (or added later with the IAM work) if that is preferable. Also it will need some tests to validate that expired cookies are handled by the built-in cookie jar or some purging logic will need adding.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+      cloudantClient.request(req, function(err, resp, data) {
+        assert.equal(err, null);
+        assert.equal(resp.statusCode, 401);
+        assert.ok(data.indexOf('"error":"unauthorized"') > -1);
+        mocks.done();
+        done();
+      });
+    });
+
+    it('returns client error 403 response on authentication failure', function(done) {
```
ricellis: I think this test is invalid.
We shouldn't have any special handling for `403` when requesting session, because I don't think it is a valid response. I think this test is valid from the perspective of getting an error code other than `200` or `401` and dealing with it correctly, but it should probably be a `5XX` for that to avoid confusion with the `403` that might be received on non-`_session` responses.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+    it('renews authentication token on 403 response', function(done) {
+      if (process.env.NOCK_OFF) {
+        this.skip();
+      }
+
+      var mocks = nock(SERVER)
+          .post('/_session', {name: ME, password: PASSWORD})
+          .reply(200, {ok: true}, MOCK_SET_COOKIE_HEADER)
+          .get(DBNAME)
+          .reply(403, {error: 'forbidden', reason: 'Forbidden'})
```
ricellis: We should **not** try a `_session` request in the case of a 403 with reason `forbidden` we should only try to request a new session in the case of a 403 with reason `credentials_expired`. We should probably have tests to cover both these scenarios instead of just the new session case.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+        })
+        .on('end', function() {
+          assert.equal(responseCount, 1);
+          assert.equal(dataCount, 1);
+          mocks.done();
+          done();
+        });
+    });
+
+    it('returns client error 403 response on authentication failure', function(done) {
```
ricellis: As above.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+    it('renews authentication token on 403 response', function(done) {
+      if (process.env.NOCK_OFF) {
+        this.skip();
+      }
+
+      var mocks = nock(SERVER)
+          .post('/_session', {name: ME, password: PASSWORD})
+          .reply(200, {ok: true}, MOCK_SET_COOKIE_HEADER)
+          .get(DBNAME)
+          .reply(403, {error: 'forbidden', reason: 'Forbidden'})
```
ricellis: As above.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+        })
+        .on('end', function() {
+          assert.equal(responseCount, 1);
+          assert.equal(dataCount, 1);
+          mocks.done();
+          done();
+        });
+    });
+
+    it('returns client error 403 response on authentication failure', function(done) {
```
ricellis: As above.

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+    it('renews authentication token on 403 response', function(done) {
+      if (process.env.NOCK_OFF) {
+        this.skip();
+      }
+
+      var mocks = nock(SERVER)
+          .post('/_session', {name: ME, password: PASSWORD})
+          .reply(200, {ok: true}, MOCK_SET_COOKIE_HEADER)
+          .get(DBNAME)
+          .reply(403, {error: 'forbidden', reason: 'Forbidden'})
```
ricellis: As above.

plugins/cookieauth.js
```diff
@@ -13,167 +13,90 @@
...
-  };
+  delete parsed.path;
+  delete parsed.pathname;
 
-  return cookieRequest;
+  this.baseUrl = u.format(parsed).replace(/\/$/, '');   // i.e. https://account.cloudant.com
+  this.refreshCookie(state, req, callback);
+};
+CookiePlugin.prototype.onResponse = function(state, response, callback) {
+  if (response.statusCode === 401 || response.statusCode === 403) {
```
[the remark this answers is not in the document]
smithsz: Ah, this could be problematic. I don't have access to the data stream from within the plugin.

plugins/cookieauth.js
smithsz: Something like this should remove expired cookies:
```js
If (this.baseUrl && req.jar.getCookies(this.baseUrl, {expire: true}).length > 0) {
  return callback(state);
}
```
Is the `refreshRequired` also needed? I was planning to set `refreshRequired = true` on 401/403. But perhaps just pruning the jar and calling `.length` on it is enough?

test/plugins/cookieauth.js
```diff
@@ -0,0 +1,984 @@
...
+
+const ME = process.env.cloudant_username || 'nodejs';
+const PASSWORD = process.env.cloudant_password || 'sjedon';
+const SERVER = `https://${ME}:${PASSWORD}@${ME}.cloudant.com`;
+const DBNAME = `/nodejs-cloudant-${uuidv4()}`;
+
+// mock cookies
+
+const MOCK_COOKIE = 'AuthSession=Y2xbZWr0bQlpcc19ZQN8OeU4OWFCNYcZOxgdhy-QRDp4i6JQrfkForX5OU5P';
+const MOCK_SET_COOKIE_HEADER = { 'set-cookie': `${MOCK_COOKIE}; Version=1; Expires=Tue, 22-Aug-2017 11:39:29 GMT; Max-Age=86400; Path=/; HttpOnly` };
```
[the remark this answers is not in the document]
smithsz: k, will remove the `Expires=` bit.

test/plugins/cookieauth.js
smithsz: So, the flow for making a request is something like this?
1. Clear expired cookies for idx _user.cloudant.com_.
2. If `cookies.length == 0` then make a `POST /_session`, on response:
   - `200`: continue
   - error || `5xx`: retry
   - default: continue _(without cookie)_
3. Make client request.

plugins/cookieauth.js
ricellis: I'm not sure how visible/cusotmizable the jar is, but as long as the thing we check for length is limited to our own `baseUrl` so we aren't counting cookies from other sources that may be in there then I think it would be OK.

test/plugins/cookieauth.js
ricellis: 1. Clear expired cookies for idx _user.cloudant.com_.
1. If `cookies.length == 0` then make a `POST /_session`, on response:
    1. `200` - store cookie, proceed with request
    1. `401` - make no more `_session` requests, continue without cookie
    1. `4xx/5xx`/error - behaves as 401 in the other libs, but I could see an argument for retrying
1. Make client request; on response:
    1. `401` || `403 {reason: credentials_expired}` - retry client request (with new `_session` request).
    1. Default pass response back

test/plugins/cookieauth.js
smithsz: As discussed, the data stream isn't currently accessible to the plugin hook so we therefore can't analyse the response body.

The problem is we must choose to retry a request before the data starts flowing. If we don't then we'll be unable to pipe to the client stream as data will have already been emitted. I'll raise an issue to capture this, although I'm not sure how to fix this at the moment.

package.json
```diff
@@ -26,7 +26,9 @@
     "async": "2.1.2",
     "cloudant-nano": "6.5.0",
     "debug": "2.2.0",
-    "request": "^2.81.0"
+    "lockfile": "1.0.3",
```
ricellis: Any reason not to put a range on this to try and keep it up to date? Seems like the sort of thing where a bug fix or optimizations/tweaks/fixes for future Node.js versions could be handy.

plugins/cookieauth.js
```diff
@@ -13,167 +13,230 @@
...
-      callback = nullcallback;
+const debug = require('debug')('cloudant:plugins:cookieauth');
+const lockFile = require('lockfile');
+const request = require('request');
+const tmp = require('tmp');
+const u = require('url');
+
+const BasePlugin = require('./base.js');
+
+function noop() {}
```
ricellis: As per #217 perhaps this noop could be removed and the lock functions extracted to somewhere common. Happy for that change to happen after merge of this and #217 if it makes things easier.

plugins/base.js
```diff
@@ -13,6 +13,12 @@
 // limitations under the License.
 'use strict';
 
+const debug = require('debug')('cloudant:plugins:base');
+const lockFile = require('lockfile');
+const tmp = require('tmp');
+
+function noop() {}
```
smithsz: I still need this noop func. If we error getting the lock then we still proceed in fetching the session cookie. This means the client is always going to call the callback.
If we success in acquiring the lock then the callback is releasing the lock, otherwise it's this noop func.

PR #211, opened 2017-08-30
Add request abort

smithsz:
## What
Implement `response.abort()`.

## How
Monkey-patch `request.clientStream.abort` with underlying `response.abort`.

## Testing
Includes additional unit test.

### DIFF
--- a/lib/clientutils.js
+++ b/lib/clientutils.js
@@ -56,7 +56,14 @@ var updateState = function(state, newState, callback) {
 
 // process state (following plugin execution)
 var processState = function(r, callback) {
-  if (r.state.abortWithResponse) {
+  if (r.abort) {
+    if (r.response) {
+      debug('client issued abort during plugin execution');
+      r.response.abort();
+    }
+    callback(new Error('Client issued abort'));
+  } else if (r.state.abortWithResponse) {
+    // => abort request and send `abortWithResponse` to client
     if (r.response) {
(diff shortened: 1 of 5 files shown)


test/fixtures/testplugins.js
```diff
@@ -43,6 +43,41 @@ NoopPlugin.prototype.onResponse = function(state, response, callback) {
   callback(state); // noop
 };
 
+// AlwaysRetry for testing
+//   - onRequest:  noop
+//   - onError:    always retries (with retry delay)
+//   - onResponse: always retries (with retry delay)
+
+function AlwaysRetry(client, opts) {
```
smithsz: I will refactor using ES2015 style classes when I rebase with [add-retry-plugins](https://github.com/cloudant/nodejs-cloudant/pull/208).

lib/clientutils.js
```diff
@@ -56,7 +56,14 @@ var updateState = function(state, newState, callback) {
 
 // process state (following plugin execution)
 var processState = function(r, callback) {
-  if (r.state.abortWithResponse) {
+  if (r.abort) {
```
ricellis: It would be nice to have some comments about the scenario each of these blocks is for so we remember down the line.

smithsz: Sure, I will add in a separate PR to prevent a nightmare rebase!

PR #213, opened 2017-08-31
Add comments to processState logic

smithsz:
## What
Add some better comments to the `processState` function.

### DIFF
--- a/lib/clientutils.js
+++ b/lib/clientutils.js
@@ -57,43 +57,46 @@ var updateState = function(state, newState, callback) {
 // process state (following plugin execution)
 var processState = function(r, callback) {
   if (r.abort) {
+    // => client has called for the request to be aborted
     if (r.response) {
       debug('client issued abort during plugin execution');
       r.response.abort();
     }
-    callback(new Error('Client issued abort'));
+    callback(new Error('Client issued abort')); // callback with error to prevent request retry
   } else if (r.state.abortWithResponse) {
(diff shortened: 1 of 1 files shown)


lib/clientutils.js
```diff
@@ -57,43 +57,50 @@ var updateState = function(state, newState, callback) {
...
-      r.state.retry = false; // => response ok || too many retries
-    }
-    if (r.eventPipe) {
-      r.eventPipe.resume(); // replay captured events
+    // => the response is "good" and can be returned to the awaiting client
+    //    OR
+    //    too many retry attempts; the "bad" response must be returned to the client
+    if (!r.state.sending) {
+      // request has not been sent, implies we are processing 'onRequest' hooks
+      if (r.state.retry) {
```
ricellis: This debug statement and flag block is present in both the `if` and `else` here, why not just do it first then do the if/else?

```js
if (r.state.retry) {
  debug('too many retry attempts');
  r.state.retry = false;
}
if (!r.state.sending) {
  // request has not been sent, implies we are processing 'onRequest' hooks
  callback(); // callback without error to proceed with executing the client's request
} else {
  // processing 'onError' or 'onResponse' hooks
  // pass response through to awaiting client
  if (r.eventPipe) {
    r.eventPipe.resume(); // replay all captured events
  }
  callback(new Error('No retry requested')); // no retry
}
```

PR #216, opened 2017-09-04
Use cloudant client

smithsz:
## What
Use the new `CloudantClient` for making all library requests.

## How
Pass the `CloudantClient` to the underlying `nano` instanance. Also, remove some legacy/redundant tests.


cloudant.js
```diff
@@ -29,67 +29,35 @@ var reconfigure = require('./lib/reconfigure.js');
...
-      plugin = options.plugin;
-    }
-  }
+  // build request client
+  var cloudantClient = new Client(options);
+  var cloudantRequest = function(req, callback) {
+    return cloudantClient.request(req, callback);
+  };
 
-  debug('Create underlying Nano instance, options=%j requestDefaults=%j', options, requestDefaults);
```
tomblench: We should keep this debug message, if it still makes sense

cloudant.js
```diff
@@ -29,67 +29,35 @@ var reconfigure = require('./lib/reconfigure.js');
...
-    theurl = reconfigure(options);
-  } else {
-    theurl = reconfigure({ url: options });
+  if (typeof options !== 'object') {
+    options = { url: options };
   }
+
+  var theurl = reconfigure(options);
   if (theurl === null) {
+    var err = new Error('Invalid URL');
```
tomblench: Is there a test case for invalid URL?

cloudant.js
```diff
@@ -29,67 +29,35 @@ var reconfigure = require('./lib/reconfigure.js');
...
-  // keep connections alive by default
-  if (requestDefaults && !requestDefaults.agent) {
-    var protocol = (theurl.match(/^https/)) ? require('https') : require('http');
-    var agent = new protocol.Agent({
-      keepAlive: true,
-      keepAliveMsecs: 30000,
-      maxSockets: 6
-    });
-    requestDefaults.agent = agent;
+  if (theurl.match(/^http:/)) {
```
tomblench: As above, wonder if we should have a test case for this

test/api.js
```diff
@@ -43,77 +43,6 @@ describe('Cloudant API', function() {
   });
 });
 
-describe('Initialization', function() {
```
ricellis: Aren't some of these tests still valid or have the things they are testing effectively been duplicated by some of your new tests? I've lost track a bit of all the changes.

cloudant.js
smithsz: This is asserted in the legacy reconfigure tests.

cloudant.js
smithsz: Yup, will add.

cloudant.js
```diff
@@ -226,6 +200,7 @@ function Cloudant(options, callback) {
   };
 
   // add top-level Cloudant-specific functions
+  nano.auth = undefined; // handled by cookie plugin
```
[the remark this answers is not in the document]
smithsz: Yeah, I guess I could leave this in. I was inclined to remove the nano authentication layer entirely. Perhaps I need to revisit the `ping` implementation too then. If the cookie plugin is used then ping is just a HEAD request, otherwise it can be the legacy fetch session thing. Also, I can put all those horrid tests back. Yay!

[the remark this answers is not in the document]
ricellis: I guess it's really a backward compatibility thing. If ping/auth worked as per nano before (without the cookie plugin) then they need to continue to do so.

PR #219, opened 2017-09-25
Tidy cloudant js

smithsz:

## What
- Change 'invalid url' message to error object
- Remove dupe `set_cors` func.
- Update documentation links.

### DIFF
--- a/cloudant.js
+++ b/cloudant.js
@@ -56,10 +56,11 @@ function Cloudant(options, callback) {
     theurl = reconfigure({ url: options });
   }
   if (theurl === null) {
+    var err = new Error('Invalid URL');
     if (callback) {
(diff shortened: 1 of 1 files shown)


cloudant.js
```diff
@@ -56,10 +56,11 @@ function Cloudant(options, callback) {
     theurl = reconfigure({ url: options });
   }
   if (theurl === null) {
+    var err = new Error('Invalid URL');
```
emlaver: Is it worth including the invalid/malformed URL value with this error message?

smithsz: I think it's okay. This all happens in the reconfigure.js code which uses `url.parse` to check for the necessary url bits (else returns null).

PR #221, opened 2017-09-29
Use cloudant client

smithsz:

## What
Use the new Cloudant client in `cloudant.js`.

## How
Pass the Cloudant client instance through to the underlying nano object.

## Testing
Includes additional/updated unit tests.

### DIFF
--- a/cloudant.js
+++ b/cloudant.js
@@ -20,41 +20,21 @@ var debug = require('debug')('cloudant:cloudant');
 var nanodebug = require('debug')('nano');
 var async = require('async');
 
-// function from the old Cloudant library to
-// parse an object { account: "myaccount", password: "mypassword"}
-// and return a URL
+const Client = require('./lib/client.js');
+
+// Parse an object (i.e. { account: "myaccount", password: "mypassword" }) and
+// return a URL.
 var reconfigure = require('./lib/reconfigure.js');
 
 // This IS the Cloudant API. It is mostly nano, with a few functions.
 function Cloudant(options, callback) {
   debug('Initialize', options);
 
-  // Save the username and password for potential conversion to cookie auth.
-  var login = reconfigure.getOptions(options);
-
-  // Convert the credentials into a URL that will work for cloudant. The
-  // credentials object will become squashed into a string, which is fine
-  // except for the .cookie option.
-  var cookie = options.cookie;
-
-  var pkg = require('./package.json');
-  var requestDefaults = {
-    gzip: true,
-    headers: {
-      // set library UA header
-      'User-Agent': `nodejs-cloudant/${pkg.version} (Node.js ${process.version})`
-    },
-    jar: false
-  };
-  var theurl = null;
-  if (typeof options === 'object') {
-    if (options.requestDefaults) {
-      requestDefaults = options.requestDefaults;
-    }
-    theurl = reconfigure(options);
-  } else {
-    theurl = reconfigure({ url: options });
+  if (typeof options !== 'object') {
+    options = { url: options };
   }
+
+  var theurl = reconfigure(options);
   if (theurl === null) {
     var err = new Error('Invalid URL');
     if (callback) {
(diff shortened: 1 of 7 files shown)


test/legacy/plugin.js
```diff
@@ -154,28 +158,6 @@ describe('promise plugin', function() {
   });
 });
 
-describe('custom plugin', function() {
```
smithsz: This is now an invalid test. All custom plugins must be a class that inherits from `BasePlugin`.

test/legacy/api.js
```diff
@@ -574,23 +568,12 @@ describe('Changes follower', function() {
     var mocks = nock(SERVER)
       .get('/' + dbName)
       .reply(200, { update_seq: '2-g1AAAADbeJzLYWBgYMlgTmGQTUlKzi9KdUhJMtUrzsnMS9dLzskvTUnMK9HLSy3JASpjSmRIsv___39WIgORGpIcgGRSPVgPI5F68liAJEMDkAJq20-8XRB9ByD6QPZlAQCMOkh4', db_name: dbName, sizes: { file: 58038, external: 8, active: 2166 }, purge_seq: 0, other: { data_size: 8 }, doc_del_count: 0, doc_count: 2, disk_size: 58038, disk_format_version: 6, compact_running: false, instance_start_time: '0' })
-      .post('/' + dbName)
-      .reply(200, {ok: true, id: docId, rev: '2-a1933e6ba0b5ac9868cd6f5adc12dc5f'})
       .filteringPath(/[?&](since=.*|feed=continuous|heartbeat=.*)/g, '') // Strip out the standard parameters: ?since &feed, &heartbeat
       .get('/' + dbName + '/_changes')
       .reply(200, '{"seq":"3-g1AAAAEzeJzLYWBgYMlgTmGQTUlKzi9KdUhJMtcrzsnMS9dLzskvTUnMK9HLSy3JASpjSmRIsv___39WBpCVCxRgT05OSjQ0SyNSe5IDkEyqB5vAnMgENsHM1CDZPDGFkH4ibchjAZIMDUAKaMl-hDtTLZMMktOMSDLlAMQUsG8ZIaYYWpqmGCVnAQDI71_n","id":"' + docId + '","changes":[{"rev":"2-a1933e6ba0b5ac9868cd6f5adc12dc5f"}]}\n');
 
-    // Make an update that will trigger the "since=now" follower.
```
smithsz: This test runs as a mock only. Therefore the flakey update_doc after 1sec logic is redundant.

cloudant.js
```diff
@@ -233,55 +209,11 @@ function Cloudant(options, callback) {
...
-      if (er) {
-        callback(er);
-      } else {
-        if (cookie) {
-          requestDefaults.headers.cookie = cookie;
-        }
-
-        callback(null, nano, pong);
-      }
+    nano.cc.addPlugins('cookieauth');
```
ricellis: So if I'm understanding this correctly, if you request the ping callback then the `cookieauth` plugin is enabled by default? I thought we were going to turn `cookieauth` on by default unless a plugin list was supplied that explicitly removed it?

test/legacy/plugin.js
ricellis: As discussed I think we should deprecate the old style custom plugin a bit more gently. I'm happy for that second path to be done in another PR if it is easier.

cloudant.js
smithsz: Yeah, the ping callback enables cookie auth. I don't like it much. But it's following the old pattern.

I haven't turned cookie auth on by default yet. It's something we can do. I wasn't too sure how a client would go about disabling it though. The cloudant client is exposed so I could add a new `removePlugins` method (to mirror `addPlugins`), i.e.
```js
const client = Cloudant({ plugins: ['retry'] });
client.cc.removePlugins('cookieauth');
```
I don't want to clutter this pr even more so I'll open a new one for this.

test/legacy/plugin.js
smithsz: Okay, this makes sense. I'll put together a PR to reinstate this test and handle legacy plugins.

PR #227, opened 2017-10-18
Support legacy plugins

smithsz:

## What
Support legacy plugins.

## How
Reinstate old code to enable users to use custom legacy plugins.

## Testing
Includes unit tests.

### DIFF
--- a/cloudant.js
+++ b/cloudant.js
@@ -15,9 +15,10 @@
 
 module.exports = Cloudant;
 
-var Nano = require('cloudant-nano');
-var debug = require('debug')('cloudant:cloudant');
-var nanodebug = require('debug')('nano');
+const async = require('async');
+const Nano = require('cloudant-nano');
+const debug = require('debug')('cloudant:cloudant');
+const nanodebug = require('debug')('nano');
 
 const Client = require('./lib/client.js');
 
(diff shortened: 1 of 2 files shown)


cloudant.js
```diff
@@ -47,15 +50,51 @@ function Cloudant(options, callback) {
...
+    debug('Using legacy custom plugin');
+    cloudantRequest = options.plugin;
+
+    if (options.requestDefaults) {
+      requestDefaults = options.requestDefaults;
+    } else {
+      requestDefaults = { gzip: true, jar: false };
+    }
+
+    // set library UA header
```
ricellis: Do we have two copies of this now?

smithsz: Yeah, this is a legacy code path. If the client specifies a legacy plugin then we avoid initialising the Cloudant client and instead substitute in their plugin. We still need to add the UA and setup the default agent config.

ricellis: It'd be nice if we could at least use the same agent/constants in both cases, but if it's not easy to put the code somewhere common then I'm ok with this.

smithsz:
I think we can pass a legacy plugin to the new `CloudantClient` where it'll be used as a direct replacement of `request`. This means we can support the use of new plugins alongside one legacy plugin. New PR to follow.

PR #228, opened 2017-10-19
Update readme

smithsz:

## What
Update `README.md` for multiple plugin support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,6 @@ This library adds documentation for the following:
 - [Cloudant Query](#cloudant-query)
 - [Cloudant Search](#cloudant-search)
 - [Cloudant Geospatial](#cloudant-geospatial)
-- [Cookie Authentication](#cookie-authentication)
 - [Advanced Features](#advanced-features)
   - [Advanced Configuration](#advanced-configuration)
   - [Pool size and open sockets](#pool-size-and-open-sockets)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,87 +220,99 @@ Cloudant({url:"companycloudant.local", username:"somebody", password:"somebody's
...
 
-#### The 'promises' Plugins
+   This plugin will automatically swap your Cloudant credentials for a cookie. It will handle the authentication for you and ensure that the cookie is refreshed when required.
+2. `promises`
 
-When initialising the Cloudant library, you can opt to use the 'promises' plugin:
+   If you'd prefer to write code in the _Promises_ style then the `promises` plugin turns each request into a _Promise_.
+3. `retry429` (or `retry`)
+
+   On occasion Cloudant's multi-tenant offering may reply with an HTTP 429 response because you've exceed the number of API requests in a given amount of time. The "retry429" plugin will automatically retry your request with an exponential back-off.
```
tomblench: retry429 should be in backticks in this para

README.md
```diff
@@ -220,87 +220,99 @@ Cloudant({url:"companycloudant.local", username:"somebody", password:"somebody's
...
+   The plugin will automatically retry request errors (e.g. connection reset errors) with an exponential back-off.
+
+_Want to use multiple plugins?_ Not a problem. Simply pass the plugins as an array instead:
 
 ```js
-var cloudant = Cloudant({url: myurl, plugin:'promises'});
+var cloudant = new Cloudant({ url: myurl, plugins: [ 'cookieauth', 'promises', 'retry5xx' ] });
 var mydb = cloudant.db.use('mydb');
+mydb.get('mydoc', function(err, data) {
+  console.log(data.toString('utf8');
```
tomblench: Maybe something like `console.log("Document contents: "+data.toString('utf8');`

README.md
```diff
@@ -220,87 +220,99 @@ Cloudant({url:"companycloudant.local", username:"somebody", password:"somebody's
...
+});
 ```
 
-Then the library will return a Promise for every asynchronous call:
+The plugins are _always_ executed in the order they are specified. Remember that all plugins are respected. If one requests a retry then it cannot be overruled by another. If two plugins request different delay times before the next retry attempt then the largest delay time is honoured.
+
+Be aware that if you don't specify any plugins then the `cookieauth` plugin will automatically be added. To disable all plugins you can pass an empty array as the plugin list, i.e. `Cloudant({ url: myurl, plugins: [] })`.
```
tomblench: Is the cookie auth plugin added if no username/password is set?

README.md
```diff
@@ -220,87 +220,99 @@ Cloudant({url:"companycloudant.local", username:"somebody", password:"somebody's
...
+#### The `cookieauth` plugin
+
+For example:
 
 ```js
-mydb.list().then(function(data) {
-  console.log(data);
-}).catch(function(err) {
-  console.log('something went wrong', err);
+var cloudant = new Cloudant({ url: myurl, plugin: 'cookieauth' });
```
tomblench: Do we need to clarify that the user/pass is in the url?

README.md
smithsz: y - it can still be added but doesn't do anything. will update docs accordingly.

README.md
smithsz: good thinking. will use 'http://user:pass@localhost:5984' instead of `myurl` perhaps?

tomblench: sounds good

PR #243, opened 2017-11-13
Support multiple plugins.

### LINKED ISSUE #44: What if credentials change mid-execution?
What if I initialize my Cloudant connection, and then credentials change in-between API calls, as mentioned in #27 

I imagine there should be a callback to the user, where they can attempt to figure out the newer credentials, and then call back to us when that is done.

Obviously, this is either a very rare situation (perhaps just throw/crash if it fails, allowing the user application to restart presumably), or else pretty advanced usage, so the documentation should reflect that (i.e. relegate this to an "advanced" section).

### LINKED ISSUE #114: Confirm and optimize session cookies
In #87 we implemented session cookies; however they are not as broadly used as they should be. Ideally, it is automatic; however with the synchronous API, I am not sure what to do.

### LINKED ISSUE #120: Switching to cookie authentication by default in this library
There may come a day when passing the username and password using Basic Authentication with every request is penalised by the server. Best practice is to call `POST /_session` to exchange your credentials for a cookie. The cookie will expire eventually, so the library will have re-authenticate before the cookie expires.

This issue proposes that we:
- on instantiation of the Cloudant object issue a `POST /session` call to get a cookie
- subsequent library calls will not use the Cloudant object's username & password, but the cookie 
- the server may send us a new cookie when making a request. This must be user in place of the existing cookie from that point on

The above tasks need only be undertaken when a username & password are supplied i.e. not for local CouchDB requests in admin party mode.

In the case where a user has a long-running application which makes requests infrequently (less frequently than the cookie interval) then the user will need to do something like

```
  setInterval(function() {
    cloudant.ping(function(err, data) { } );
  }, 1000 * 60 * 60);
```

to ensure that the cookie stays fresh. 

---

The idea here is that by default, this library will always exchange authentication credentials for cookies and this will happen transparently for the user.

### LINKED ISSUE #175: Support several plugins at the same time
The library supports *request* plugins (https://github.com/cloudant/nodejs-cloudant#request-plugins) but only one can be specified.

What if I want to enable the retry and promises?

smithsz:

## What
Support multiple plugins.

## How
Run plugins using a request interceptor model.

## Testing
Includes additional tests.

## Issues
Fixes #175. 
Fixes #120.
Fixes #114.
Fixes #44.


CHANGES.md
```diff
@@ -1,4 +1,8 @@
-# UNRELEASED
+# 2.0.0 (UNRELEASED)
+- [NEW] Added API for upcoming IBM Cloud Identity and Access Management support
+  for Cloudant on IBM Cloud. Note: IAM API key support is not yet enabled in the
+  service.
+- [NEW] Support multiple plugins.
```
tomblench: Should we have a link to the documentation? I understand that the upgrade implications are minimal and that existing use of one plugin will continue to work seamlessly, is that right and should that be made more obvious?

api-migration.md
```diff
@@ -0,0 +1,28 @@
...
+version to another. The section titles state the specific versions between
+which the change was made; often one can migrate from earlier versions than
+the one specified on the left of the title to the new version specified
+on the right.
+
+Feel free to file bugs if the instructions don't work for you; we'll use
+these to improve the documentation and we'll obviously also try to help
+you through the process.
+
+## 1.10.0 → 2.0.0
```
tomblench: I think this can change from 1.x to 2.x and tweak the text above to re-assure the user that breaking changes typically (hopefully only) occur on major version changes

api-migration.md
```diff
@@ -0,0 +1,28 @@
+# Migrating to new APIs
+
+This document covers migrating your code when a nodejs-cloudant release has
+breaking API changes. Each section covers migrating from one particular
+version to another. The section titles state the specific versions between
+which the change was made; often one can migrate from earlier versions than
+the one specified on the left of the title to the new version specified
+on the right.
+
+Feel free to file bugs if the instructions don't work for you; we'll use
```
tomblench: This sounds like you don't have much confidence in the instructions. I'm not sure of a better wording but customers generally don't hold back in reporting bugs when things break. Do we need this para?

smithsz: This was all stolen from CTD. I think you're right. I'll drop it.

api-migration.md
smithsz: Will do.

PR #252, opened 2017-11-30
Include piping note in api-migration.md

smithsz:

## What
Include piping note in `api-migration.md`.

### DIFF
--- a/api-migration.md
+++ b/api-migration.md
@@ -15,6 +15,28 @@ The library continues to support legacy plugins. They can be used in conjunction
 with new plugins. Your plugins list may contain any number of new plugins but
 only ever one legacy plugin.
 
+The `cloudant.request` object is a pass through stream and cannot be used inside
+an event hook function. For example:
+
+```js
+var cloudant = new Cloudant({ url: myUrl, plugins: [ 'retry429', 'retryerror' ] });
+
+// make a custom request
+var r = cloudant.request({ path: '_all_dbs' })
+  .on('response', function(resp) {
+    if (resp.statusCode === 200) {
+      r.pipe(process.stdout); // this doesn't work!
+      resp.pipe(process.stdout); // this does work!
+    }
+  });
+
+r.pipe(process.stdout); // this does work!
+```
+
+The final response isn't returned to the client until all the plugin hooks have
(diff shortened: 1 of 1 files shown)


api-migration.md
```diff
@@ -15,6 +15,27 @@ The library continues to support legacy plugins. They can be used in conjunction
 with new plugins. Your plugins list may contain any number of new plugins but
 only ever one legacy plugin.
 
+It's important to use the response object returned by the event hook when
```
emlaver: What about merging the two sentences to make it clear that only the response object is available when piping and not the request object.  Something like: `It's important to note that cloudant.request object is a pass through stream and only the response object returned by the event hook is usable when piping`

smithsz: Thanks, I've had another bash at this. Hopefully it makes a little more sense.

api-migration.md
```diff
@@ -15,6 +15,29 @@ The library continues to support legacy plugins. They can be used in conjunction
 with new plugins. Your plugins list may contain any number of new plugins but
 only ever one legacy plugin.
 
+It's important to note that the `cloudant.request` object is a pass through
+stream and isn't always suitable for piping. However, the response object
```
emlaver: `isn't always suitable` - should we explain when it is suitable? Or is it better to just say something like `isn't a reliable solution for piping`.

api-migration.md
```diff
@@ -15,6 +15,29 @@ The library continues to support legacy plugins. They can be used in conjunction
...
+returned by the event hook can always be piped. This is because it's not
+guaranteed that any given response will be returned to the client until all the
```
emlaver: `This is because...` - this sentence is awkward.  Is there a way to simply it with the previous sentence? Maybe something like:
`If piping is required, use the response object returned by the event hook since the client will return a complete response once all 'onResponse' hooks have been executed.`

api-migration.md
smithsz: yeah - I guess I want to clarify that it can't be used in an event hook.

api-migration.md
smithsz: Again, it can only be used outside the event hook. Otherwise use the response object returned by the event hook. Will try and clarify.

PR #253, opened 2017-12-05
139 fix security asymmetry

### LINKED ISSUE #139: get_security / set_security asymmetrical
`db.get_security()` seems to return the entire `_security` record as per https://cloudant.com/for-developers/faq/auth/ 

``` js
{
  "_id": "security", 
  "cloudant": {
    "exampleuser1": [
      "_reader", 
      "_writer"
    ]
  }
}
```

however, `db.set_security(x)` does a `POST` to `_security`  with `{ cloudant: x }` 

so, if i'm trying to add user permissions for something, I can't do  `get_security()`, mutate the value with `x.user=['_reader']`, and just call `set_security(x)`. I end up with extra `cloudant: {}` objects in the security record that way.

smithsz:

## What
Fix `get_security`/`set_security` asymmetry.

## Testing
Updates unit tests.

## Issues
Fixes #139.

### DIFF
--- a/cloudant.js
+++ b/cloudant.js
@@ -91,10 +91,21 @@ function Cloudant(options, callback) {
 
    // https://console.bluemix.net/docs/services/Cloudant/api/authorization.html#modifying-permissions
     var set_security = function(permissions, callback) { // eslint-disable-line camelcase
-      var path = '_api/v2/db/' + encodeURIComponent(db) + '/_security';
-      return nano.request({ path: path,
+      var body = permissions;
+      var prefix = '_api/v2/db/'; // use `/_api/v2/<db>/_security` endpoint
+
+      if (permissions['couchdb_auth_only']) {
+        // https://console.bluemix.net/docs/services/Cloudant/api/authorization.html#using-the-_users-database-with-cloudant-nosql-db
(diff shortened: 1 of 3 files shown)


cloudant.js
```diff
@@ -92,9 +92,11 @@ function Cloudant(options, callback) {
    // https://console.bluemix.net/docs/services/Cloudant/api/authorization.html#modifying-permissions
     var set_security = function(permissions, callback) { // eslint-disable-line camelcase
       var path = '_api/v2/db/' + encodeURIComponent(db) + '/_security';
-      return nano.request({ path: path,
+      return nano.request({
+        path: path,
         method: 'put',
-        body: {cloudant: permissions} }, callback);
```
ricellis: Is it worth doing something like this so people can get it right either way:
```js
if permissions.cloudant {
  body = permissions
} else {
  body = {cloudant: permissions}
}
```

I'm not sure we should add the complexity given that the doc makes it clear what kind of object is required, but am just thinking "aloud".

smithsz: Unsure. Would this still work with `couchdb_auth_only`? You'd need to specify a `cloudant` key I guess.

Issue #258, opened 2018-01-04
Support streaming a file to a PUT or POST request


smithsz:
_Example Failure:_ 
https://github.com/cloudant/couchbackup/commit/59676c6253494485c4d73f0a019656e9bff9a670

Also see failing build, https://cloudant-sdks-jenkins.swg-devops.com:8443/blue/organizations/jenkins/couchbackup/detail/update-cloudant-version/2/pipeline.

smithsz:
Fixed.

PR #260, opened 2018-01-10
Support piping of request payload

smithsz:

## What
Support payload piping, for example:
```js
fs.createReadStream('keys.json').pipe(cloudant.request({ path: 'my_db/_all_docs', method: 'POST' }).pipe(process.stdout)
```
_Note:_ Additional work is required if these types of request are to be retried since currently the request stream may only be read once. This will be handled in a future PR.

## How
`cloudant.request` now returns a Duplex type which implements both Readable and Writable interfaces allowing both the request and response to be streamed.

## Testing

## Issues
See #258.

### DIFF
--- a/lib/eventrelay.js
+++ b/lib/eventrelay.js
@@ -1,4 +1,4 @@
-// Copyright © 2017 IBM Corp. All rights reserved.
+// Copyright © 2017, 2018 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


test/client.js
```diff
@@ -438,6 +439,34 @@ describe('CloudantClient', function() {
...
+          url: SERVER + DBNAME + '/_all_docs',
+          auth: { username: ME, password: PASSWORD },
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' }
+        };
+
+        readable.pipe(cloudantClient.request(options, function(err, resp, data) {
+          assert.equal(err, null);
+          assert.equal(resp.statusCode, 200);
+          assert.ok(data.indexOf('"key":"doc1","error":"not_found"') > -1);
```
tomblench: Should there be an assert that `'{"keys":["doc1"]}'` was written to the mock server?

test/client.js
```diff
@@ -469,6 +498,35 @@ describe('CloudantClient', function() {
...
```
tomblench: As above

test/client.js
```diff
@@ -533,6 +591,37 @@ describe('CloudantClient', function() {
...
```
tomblench: As above

test/client.js
smithsz: yeah I guess. I couldn't be bothered to add the @before func that creates the doc so was testing for the not_found. but since you've mentioned it...

smithsz: ah, I've pinged  André Rouél it seems. apols. he must get that a lot!

smithsz: Fixed in bf83319.

test/client.js
smithsz: Fixed in bf83319.

test/client.js
smithsz: Fixed in bf83319.

test/client.js
```diff
@@ -438,13 +456,44 @@ describe('CloudantClient', function() {
...
+            .post(DBNAME + '/_all_docs', function(body) {
+              assert.deepEqual(body, { keys: [ 'doc1' ] });
+              return true;
+            })
+            .reply(200, { rows: [{ key: 'doc1', value: { rev: '1-xxxxxxxx' } }] });
+
+        var cloudantClient = new Client({ plugins: [] });
+        assert.equal(cloudantClient._plugins.length, 0);
+
+        var readable = new stream.Readable();
```
ricellis: you could also do:
```
var readable = new stream.PassThrough();
readable.end('{"keys":["doc1"]}');
```
I guess it saves a line, but it's not much of an optimisation.

lib/eventrelay.js
```diff
@@ -26,109 +26,58 @@ class EventRelay {
...
+      self._target = target;
+      self.setSource(source);
     }
 
-    self.setSource(source || {});
-    self._target = target;
-
-    self._disablePiping = false;
-    self._eventsStash = [];
+    self._enablePiping = true;
```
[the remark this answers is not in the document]
smithsz: nope, will remove.

test/client.js
smithsz: wil change.

PR #266, opened 2018-01-24
Scope plugin config

### LINKED ISSUE #255: Namespace plugin configuration
Ensure configuration is only visible to the intended plugin.

Example:
```
{
  url: myurl,
  plugins: [ 'retryerror', 'retry5xx' ],
  retryerror: { maxAttempt: 5 },
  retry5xx: { maxAttempt: 3 }
 }
```

smithsz:

## What
Scope plugin configuration, for example:

```js
var cloudant = new Cloudant({ url: myUrl, iamApiKey: 'xxxxxxxxxx', plugin: 'iamauth' });
```
_becomes..._

```js
var cloudant = new Cloudant({ url: myUrl, plugin: { iamauth: { iamApiKey: 'xxxxxxxxxx' } } });
```

Also, removes support for legacy configuration flags.

## How
Gather configuration on client construction, applying defaults as required. Don't allow configuration changes at request time. Don't pass plugin configuration around in the `state`.

## Testing
Updates affected unit tests.

## Issues
Fixes #255.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
 
 ### Request Plugins
 
-This library can be used with one (or more) of the following `request` plugins:
+The library is easily extendable via the use of plugins. They provide the ability to intercept a request:
+1. Before the request is submitted to the server.
+2. After the response headers are received.
+3. If the underlying HTTP client emits an `error` event.
+
+Plugins can be used to modify an outgoing request, edit an incoming response or even retry a request entirely.
+
+#### Plugin Configuration
+
+The following global configurations can be defined to affect the behavior of all plugins:
+-	`maxAttempt` (or `retryAttempts`) - The maximum number of times the request will be attempted _(default: 3)_.
+-	`retryInitialDelayMsecs` (or `retryTimeout`) - The initial retry delay in milliseconds _(default: 500)_.
+-	`retryDelayMultiplier` - The multiplication factor used for increasing the timeout after each subsequent attempt _(default: 2)_.
+
+Note that `retryInitialDelayMsecs` and `retryDelayMultiplier` can be overriden by plugin specific configuration. This is achieved by passing the plugin (and its associated configuration) as an object.
+
+For example:
+```js
+var cloudant = new Cloudant({ url: myurl, retryDelayMultiplier: 3, plugins: [ 'retryerror', { retry5xx: { retryDelayMultiplier: 4 } } ]);
+var mydb = cloudant.db.use('mydb');
+```
+
+`maxAttempt` can _not_ be overridden by plugin specific configuration.
+
+#### The Plugins
 
 1. `cookieauth`
 
    This plugin will automatically exchange your Cloudant credentials for a cookie. It will handle the authentication and ensure that the cookie is refreshed as required.
 
+   For example:
+   ```js
+   var cloudant = new Cloudant({ url: 'http://user:pass@localhost:5984', plugin: 'cookieauth' });
+   var mydb = cloudant.db.use('mydb');
+   mydb.get('mydoc', function(err, data) {
+     console.log(`Document contents: ${data.toString('utf8')}`);
+   });
+   ```
+
+   The plugin will transparently call `POST /_session` to exchange your credentials for a cookie before proceeding with the document fetch.
+
+   Note that all subsequent requests made using this client will also use cookie authentication. The library will automatically refresh the cookie on any `401` or `403` response.
+
+   If you don't specify a username and password during the client construction then cookie authentication is disabled.
+
 2. `iamauth`
 
    IBM Cloud Identity & Access Management enables you to securely authenticate users and control access to all cloud resources consistently in the IBM Bluemix Cloud Platform.
+
    This plugin will automatically exchange your IAM API key for a token. It will handle the authentication and ensure that the token is refreshed as required.
-   The production IAM token service at https://iam.bluemix.net/oidc/token is used by default. You can set `iamTokenUrl` in your client configuration to override this.
+
+   The production IAM token service at https://iam.bluemix.net/oidc/token is used by default. You can set `iamTokenUrl` in your plugin configuration to override this.
+
+   For example:
+   ```js
+   var cloudant = new Cloudant({ url: 'http://user:pass@localhost:5984', plugin: { iamauth: { iamApiKey: 'xxxxxxxxxx', iamTokenUrl: 'https://my.iam.token.url.com/token' } } });
+   var mydb = cloudant.db.use('mydb');
+   mydb.get('mydoc', function(err, data) {
+     console.log(`Document contents: ${data.toString('utf8')}`);
+   });
+   ```
+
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
...
+Plugins can be used to modify an outgoing request, edit an incoming response or even retry a request entirely.
+
+#### Plugin Configuration
+
+The following global configurations can be defined to affect the behavior of all plugins:
+-	`maxAttempt` - The maximum number of times the request will be attempted _(default: 3)_.
+-	`retryInitialDelayMsecs` - The initial retry delay in milliseconds _(default: 500)_.
+-	`retryDelayMultiplier` - The multiplication factor used for increasing the timeout after each subsequent attempt _(default: 2)_.
+
+Note that `retryInitialDelayMsecs` and `retryDelayMultiplier` can be overriden by plugin specific configuration. This is achieved by passing the plugin (and its associated configuration) as an object.
```
smithsz: I'm planning to combine all the retry plugins, i.e.

```
plugin: [ { retry: { retryDelayMultiplier: 4, retryStatusCodes: [ 429, 500, 501 ], retryError: true } } ]
```

These settings can then be removed from the global configuration scope and be plugin specific.

lib/client.js
```diff
@@ -50,28 +49,98 @@ class CloudantClient {
...
 
     // initialize the internal client
     self._initClient(client);
 
     // add plugins
-    self.addPlugins(plugins);
+    self._addPlugins(plugins);
+  }
+
+  _addPlugins(plugins) {
```
smithsz: This is now private. Although new plugins could be added after construction, the legacy plugins can't as they overwrite the underlying request client being used.

README.md
```diff
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
...
-var cloudant = new Cloudant({ url: myurl, plugin: 'retry429' });
+var cloudant = new Cloudant({ url: myurl, plugins: [ 'cookieauth', 'promises', { retry5xx: { retryDelayMultiplier: 4 } } ] });
 var mydb = cloudant.db.use('mydb');
+mydb.get('mydoc', function(err, data) {
+  console.log(`Document contents: ${data.toString('utf8')}`);
+});
 ```
 
-Then use the Cloudant library normally. You may also opt to configure the retry parameters:
+Note that `plugins` is an alias for `plugin`. Both will accept single or multiple plugin configurations. They are interchangeable.
```
ricellis: Need we bother with this alias given that we've got other breaking changes? I guess you're allowing backwards compatibility for cases like `plugin: 'retry'` but given the configuration is going to move we won't be fully backwards compatible so maybe it is more confusing to provide half of it?

lib/client.js
```diff
@@ -50,28 +49,98 @@ class CloudantClient {
...
-          } else {
-            debug('Using legacy plugin.');
-            self.useLegacyPlugin = true;
-            client = plugin; // use legacy plugin as client
-            return;
-          }
+        if (typeof plugin !== 'function' || plugin.pluginVersion >= 2) {
+          plugins.push(plugin);
+        } else if (self.useLegacyPlugin) {
+          throw new Error('Using multiple legacy plugins in not supported');
```
ricellis: `in` -> `is`

and perhaps we should say `permitted` rather than `supported` given the latter may feel a bit like "you can try if you want, but you're on your own", but the former is the more correct "we won't allow you to use this configuration".

lib/client.js
```diff
@@ -50,28 +49,98 @@ class CloudantClient {
...
+
+      if (typeof p === 'function') {
+        debug(`Found custom plugin: '${p.id}'`);
+        Plugin = p;
+        cfg = {};
+      } else {
+        var plugin = {};
+
+        // expected: { 'pluginName': { 'configKey1': 'configValue1', ... } }
+        if (typeof p !== 'object') {
```
ricellis: At the moment this assumes a string for non-object, but what if someone accidentally specifies `1` or `true` as a plugin?

README.md
```diff
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
...
-mydb.get('mydoc', function(err, data) {
-  console.log(`Document contents: ${data.toString('utf8')}`);
-});
-```
+   The production IAM token service at https://iam.bluemix.net/oidc/token is used by default. You can set `iamTokenUrl` in your plugin configuration to override this.
 
-The plugins are _always_ executed in the order they are specified. Remember that all plugins are respected. If one requests a retry then it cannot be overruled by another. If two plugins request different delay times before the next retry attempt then the largest delay time is honoured.
+   For example:
+   ```js
+   var cloudant = new Cloudant({ url: 'http://user:pass@localhost:5984', plugin: { iamauth: { iamApiKey: 'xxxxxxxxxx', iamTokenUrl: 'https://my.iam.token.url.com/token' } } });
```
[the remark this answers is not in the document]
ricellis: I also think people are likely to copy/paste config from the `README` so we should remove the iamTokenUrl to encourage people to accept the default. We have a statement of how to override it, that should be sufficient.

README.md
```diff
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
...
+#### The Plugins
 
 1. `cookieauth`
 
    This plugin will automatically exchange your Cloudant credentials for a cookie. It will handle the authentication and ensure that the cookie is refreshed as required.
 
-2. `iamauth`
+   For example:
+   ```js
+   var cloudant = new Cloudant({ url: 'http://user:pass@localhost:5984', plugin: 'cookieauth' });
```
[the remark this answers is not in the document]
smithsz: Changed all URLs in README.md to `https://user:pass@examples.cloudant.com`.

README.md
```diff
@@ -230,107 +230,137 @@ The `ping()` function is the only exception to this rule. It does not return hea
...
+   This plugin will automatically exchange your IAM API key for a token. It will handle the authentication and ensure that the token is refreshed as required.
 
-```js
-var cloudant = new Cloudant({ url: myurl, plugins: [ 'cookieauth', 'promises', 'retry5xx' ] });
-var mydb = cloudant.db.use('mydb');
-mydb.get('mydoc', function(err, data) {
-  console.log(`Document contents: ${data.toString('utf8')}`);
-});
-```
+   The production IAM token service at https://iam.bluemix.net/oidc/token is used by default. You can set `iamTokenUrl` in your plugin configuration to override this.
```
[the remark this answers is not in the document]
smithsz: Ah, yes. Planning to merge this one first and rebase.

README.md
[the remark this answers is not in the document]
smithsz: Removed `iamTokenUrl` configuration.

README.md
smithsz: Removed legacy `plugin` parameter in 646f7bfffee685e143d53c37a26622229a26b6eb.

lib/client.js
smithsz: Done.

lib/client.js
smithsz: It will throw `Failed to load plugin`.

Added some additional tests in 4c2b4503de1728ff5b5cda03e97eca9bbb4931ab.

Issue #220, opened 2017-09-25
TypeScript support


rafaelyates:
Hi everyone, it would be nice to have the typescript support in the library, currently there's no types/cloudant , thanks

ericmillin:
I came here to post this very issue. There are already typings for `nano`, so should be fairly simple, though would need to add the Promises as return types, I guess: https://github.com/mateusnroll/typed-nano/blob/master/lib/nano.d.ts

tomblench:
I'm afraid this isn't a priority for us at the moment.

We have some work in progress to enable promises on nodejs-cloudant which may make future typescript integration easier.

garrensmith:
I've created a PR for the types https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23762

smithsz:
Release 2.1.0 has basic TypeScript support.

I've created https://github.com/cloudant/nodejs-cloudant/issues/289 to support `Promise` return types in the declaration files.

PR #280, opened 2018-02-20
add type definitions for cloudant

garrensmith:

## What

This adds typescript definitions. With this PR, any code base that depends on Cloudant and uses typescript will get some predefined definitions.

## How

I created a type definition based off of the nano definitions. I also added a test to be run.

## Testing

I have added `dslint` which is the testing tool recommended for testing type definitions. I have also tested it against an internal Cloudant project that uses typescript and this library.

## Issues

https://github.com/cloudant/nodejs-cloudant/issues/220


garrensmith:
Is there any specific reason to support node 4 in this library? That version has long been retired.

ricellis:
We've been trying to test with the LTS versions + current release.
Node 4.x is still in Maintenance LTS until the end of April.

types/index.d.ts
```diff
@@ -0,0 +1,128 @@
+import * as nano from 'nano';
```
smithsz: Add the copyright header to all new files please.

types/tests.ts
```diff
@@ -0,0 +1,75 @@
...
+const cors: cloudant.CORS = {
+  enable_cors: true,
+  allow_credentials: true,
+  origins: ["*"]
+};
+
+instance.set_cors(cors, (error, data) => {});
+
+const virtualHost: cloudant.VirtualHost = {
+  host: 'www.host.com',
```
smithsz: Use an IANA reserved domain like example.com here.

garrensmith:
@smithsz thanks for the review. I've updated the PR. I'm not sure what you want to do about the node 4 failing errors though?

types/index.d.ts
```diff
@@ -0,0 +1,142 @@
...
+// limitations under the License.
+
+import * as nano from 'nano';
+import { Request, CoreOptions } from "request";
+
+declare function cloudant(
+    config: cloudant.Configuration | string
+  ): cloudant.ServerScope | cloudant.DatabaseScope;
+
+declare namespace cloudant {
```
ricellis: I'm a bit confused about why this is a `namespace` since Node.js uses modules. Honestly I don't know enough about Node.js packaging and TypeScript to know what is right or not, but I'm going [from this statement](http://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html):
>To reiterate why you shouldn’t try to namespace your module contents, the general idea of namespacing is to provide logical grouping of constructs and to prevent name collisions. Because the module file itself is already a logical grouping, and its top-level name is defined by the code that imports it, it’s unnecessary to use an additional module layer for exported objects.

Again I know next to nothing about TypeScript and have had about 5 minutes with the docs, but I feel like this should be following the pattern of the [module template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) - what am I misunderstanding here?

package.json
```diff
@@ -23,6 +23,8 @@
     "database"
   ],
   "dependencies": {
+    "@types/nano": "^6.4.5",
+    "@types/request": "^2.47.0",
```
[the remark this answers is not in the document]
garrensmith: I'm not 100% sure on that. But the tests I  have added should lint all the types we are using and check that they are compatible. But I don't think it would catch everything. We can lock those types down to a specific version

package.json
```diff
@@ -32,30 +34,33 @@
...
-    "test": "eslint --ignore-path .eslintignore . && mocha && nsp check",
+    "test": "eslint --ignore-path .eslintignore . && mocha && npm run dtslint && nsp check",
     "test-verbose": "env DEBUG='*,-mocha:*' npm run test",
     "test-live": "NOCK_OFF=true mocha",
-    "test-live-verbose": "env DEBUG='*,-mocha:*' npm run test-live"
+    "test-live-verbose": "env DEBUG='*,-mocha:*' npm run test-live",
+    "dtslint": "dtslint types"
   },
   "main": "./cloudant.js",
+  "types": "types",
```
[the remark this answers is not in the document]
garrensmith: It shouldn't matter. I took that style from the [dtslint guide](https://github.com/Microsoft/dtslint#add-types-for-a-library-not-on-definitelytyped)

garrensmith:
@smithsz your changes look good +1

ricellis:
@garrensmith @smithsz I'm basically OK with this, but I think we need something in the readme about how it won't work with promises, or how to make it work with promises (e.g. disabling).

PR #296, opened 2018-03-21
Support Cloudant query when using promises request plugin

### LINKED ISSUE #292: Getting indices with index() errors with 'promises' plugin enabled
Cloudant version: 2.1.0
Node version: 9.8.0

Code Sample:
```
import * as Cloudant from '@cloudant/cloudant';
const cloudantInstance = Cloudant({
  url: cloudantConnectionUrl,
  plugins: [
    'cookieauth',
    'promises',
  ],
});
const myDb = cloudantInstance.db.use('my-db');
const indizes = await myDb.index().catch(error => console.log(error));
```

The error logged in the last line results in a 400 'invalid UTF-8 JSON' from an otherwise perfectly operating Cloudant instance. Not using the 'promises' plugin with the same Cloudant instance also works fine.

The error message and a quick look into the code leads me to believe that this library incorrectly tries to create an index without an input object instead of getting the indices from the database, because the `index` function (file: cloudant.js, line: 114) checks for a callback function argument, which isn't and shouldn't be available in 'promises' mode.

smithsz:

## What
Support Cloudant query when using promises request plugin.

## How

List all Cloudant NoSQL DB Query indexes:
```js
db.index().then((d) => { console.log(d); });
```

Creating an index:
```js
db.index(definition).then((d) => { console.log(d); })
```

## Testing

## Issues
Fixes #292.

### DIFF
--- a/test/issues/292.js
+++ b/test/issues/292.js
@@ -1,0 +1,107 @@
+// Copyright © 2018 IBM Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* global describe it before after */
+'use strict';
+
+const assert = require('assert');
+const Client = require('../../lib/client.js');
+const Cloudant = require('../../cloudant.js');
+const nock = require('../nock.js');
+const uuidv4 = require('uuid/v4'); // random
+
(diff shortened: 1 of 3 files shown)


test/issues/292.js
```diff
@@ -0,0 +1,107 @@
...
+const Cloudant = require('../../cloudant.js');
+const nock = require('../nock.js');
+const uuidv4 = require('uuid/v4'); // random
+
+const ME = process.env.cloudant_username || 'nodejs';
+const PASSWORD = process.env.cloudant_password || 'sjedon';
+const SERVER = `https://${ME}.cloudant.com`;
+const DBNAME = `nodejs-cloudant-${uuidv4()}`;
+
+describe('Issue #292', function() {
```
emlaver: Wouldn't it be better to put these tests under a file named like `index_with_promise_tests.js`? This way we could put any future index tests in the same location.

smithsz: These tests are specific to the issue. I'd prefer index tests to live in the `cloudant.js` test file. Note there are no index tests currently.

emlaver: Ok sounds good.

PR #307, opened 2018-05-07
Added support for IAM API key when initializing client with VCAP_SERVICES

emlaver:

## What

Added support for IAM API key when creating a Cloudant client using `vcapServices`:
```javascript
var VCAP_SERVICES = '{"cloudantNoSQLDB": [{"credentials": {"apikey": "api123key", "username": "bluemix-user"}}]}'
var cloudant = Cloudant({ vcapServices: JSON.parse(VCAP_SERVICES) });
```

## How

- Updated `reconfigure.js` to return a dictionary object with either a) url with legacy creds or b) IAM with url
- If IAM key exists in vcap services, add `IAMPlugin` to client
- Added tests to assert vcap services IAM key in `reconfigure.js`

## Testing

Added tests in `reconfigure` for asserting IAM key and URL when creating client with single and multiple vcap services.

### DIFF
--- a/cloudant.js
+++ b/cloudant.js
@@ -1,4 +1,4 @@
-// Copyright © 2015, 2017 IBM Corp. All rights reserved.
+// Copyright © 2015, 2018 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -33,17 +33,19 @@ function Cloudant(options, callback) {
     options = { url: options };
   }
 
-  var theurl = reconfigure(options);
-  if (theurl === null) {
+  var creds = reconfigure(options);
+  if (!creds || creds.outUrl === null) {
     var err = new Error('Invalid URL');
     if (callback) {
       return callback(err);
     } else {
       throw err;
     }
+  } else {
+    options.creds = creds;
   }
 
-  if (theurl.match(/^http:/)) {
+  if (creds.outUrl.match(/^http:/)) {
     options.https = false;
   }
 
@@ -53,7 +55,7 @@ function Cloudant(options, callback) {
     return cloudantClient.request(req, callback);
   };
 
-  var nanoOptions = { url: theurl, request: cloudantRequest, log: nanodebug };
+  var nanoOptions = { url: creds.outUrl, request: cloudantRequest, log: nanodebug };
   if (options.cookie) {
     nanoOptions.cookie = options.cookie; // legacy - sets 'X-CouchDB-WWW-Authenticate' header
   }
--- a/lib/reconfigure.js
+++ b/lib/reconfigure.js
@@ -1,4 +1,4 @@
-// Copyright © 2015, 2017 IBM Corp. All rights reserved.
+// Copyright © 2015, 2018 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


cloudant.js
```diff
@@ -33,8 +33,8 @@ function Cloudant(options, callback) {
     options = { url: options };
   }
 
-  var theurl = reconfigure(options);
-  if (theurl === null) {
+  var creds = reconfigure(options);
+  if (!creds && creds.outUrl === null) {
```
smithsz: Should this be `||`?

cloudant.js
```diff
@@ -43,17 +43,21 @@ function Cloudant(options, callback) {
...
 
-  if (theurl.match(/^http:/)) {
+  if (creds.outUrl.match(/^http:/)) {
     options.https = false;
   }
 
   debug('Creating Cloudant client with options: %j', options);
+  // If IAM key found in VCAP, add IAM plugin
+  if (creds && creds.iamApiKey) {
+    options.plugins = { iamauth: { iamApiKey: creds.iamApiKey } };
```
smithsz: Thinking about this, are we overwriting any plugins specified by the user here? I think we should be updating the plugin map if it exists.

emlaver: Yes, it probably would have overwritten existing plugins.  I've fixed this in e9365b1.  I've successfully tested the new logic with the the two client constructors below:
```
var cloudant = Cloudant({ vcapServices: JSON.parse(config)});
```
```
var cloudant = Cloudant({ vcapServices: JSON.parse(config), plugins: [ 'promises', { retry: { retryDelayMultiplier: 4 } } ] });
```

cloudant.js
emlaver: Yes, fixed in e9365b1.

lib/reconfigure.js
```diff
@@ -79,7 +79,11 @@ module.exports = function(config) {
     for (var i = 0; i < cloudantServices.length; i++) {
       if (typeof config.instanceName === 'undefined' || cloudantServices[i].name === config.instanceName) {
         var credentials = cloudantServices[i].credentials;
-        if (credentials && credentials.url) {
+        if (credentials && credentials.apikey && credentials.username) {
+          outCreds.iamApiKey = credentials.apikey;
+          outUrl = 'https://' + encodeURIComponent(credentials.username) + '.cloudant.com';
```
ricellis: I think we should set the URL from the `host` property of the VCAP data (for both IAM and legacy auth cases). That will work better for cases where the name isn't guaranteed to be `cloudant.com`.

cloudant.js
```diff
@@ -43,17 +43,26 @@ function Cloudant(options, callback) {
...
   }
 
-  if (theurl.match(/^http:/)) {
+  if (creds.outUrl.match(/^http:/)) {
     options.https = false;
   }
 
   debug('Creating Cloudant client with options: %j', options);
+  // If IAM key found in VCAP, add IAM plugin
+  if (creds && creds.iamApiKey) {
```
ricellis: In `client.js` the default `cookieauth` plugin is added. Do you think it makes sense to align the approaches a bit more and add the (default) `cookieauth` here in a similar way to IAM? I'm not sure how much pain that would cause, but it might make it easier to follow if defaults are added in the same way and place.

lib/reconfigure.js
emlaver: Ok, sounds good, I'll update this.

cloudant.js
emlaver: @ricellis I think they are both about the same.  I can move the default cookie auth to `cloudant.js`.

lib/reconfigure.js
emlaver: Updated in 93c3da0.

PR #311, opened 2018-05-14
Removed addition of statusCode to response objects

### LINKED ISSUE #308: Return values when using Promises plugin not correct
Please include the following information in your ticket.

- Cloudant (nodejs-cloudant) version(s) that are affected by this issue: Tested on 2.1.0 and 2.2.0
- Node version : v6.11.0
- A small code sample that demonstrates the issue.

This may be user error, but I'm struggling with the promises plugin.  Return values from a simple get seem to include a statusCode field in the body:

~~~~
const Cloudant = require('@cloudant/cloudant');
const cloudant = new Cloudant({ url: CDB_URL, plugins: [ 'promises', { retry: { retryDelayMultiplier: 4 } } ] });
    let db = cloudant.use("someDb");
    db.get("someId")
        .then((res) => {
            console.log(res)
       });
~~~~

The output would be something like:

{ endpoints: [],
  agentc:
   { switch:
      { certs: [Object],
        url: 'blah' } },
  statusCode: 200 }

which includes a statusCode.

The effects are worse when using attachments as db.attachment.get does not return a Byte stream, but an Object which includes a statusCode.

If I use regular callbacks, all works as expected.  Hopefully this is simple error on my part, but if someone could help me out that would be great.

smithsz:

## What
Removed addition of `statusCode` to response objects.

## Testing
Update existing unit test to assert object key length.

## Issues
Fixes #308.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # UNRELEASED
+- [FIXED] Removed addition of `statusCode` to response objects returned by promises.
 - [IMPROVED] Added support for IAM API key when initializing client with VCAP_SERVICES environment variable.
 
 # 2.2.0 (2018-04-30)
--- a/lib/client.js
+++ b/lib/client.js
@@ -205,7 +205,6 @@ class CloudantClient {
           if (data) {
             try {
               data = JSON.parse(data);
-              data.statusCode = response.statusCode;
             } catch (err) {}
           } else {
             data = { statusCode: response.statusCode };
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -1,4 +1,5 @@
 # UNRELEASED
+- [FIXED] Removed addition of `statusCode` to response objects.
```
ricellis: append `returned by promises` ?

PR #324, opened 2018-06-13
Handle cases when username, password, and url exist during client creation

### LINKED ISSUE #323: Properly handle authentication when providing username, password, and url
- Cloudant (nodejs-cloudant) version: 2.3.0

We currently don't handle cases in `reconfigure.js` when constructing a client that contains username, password, and url.  One of these cases would be for authenticating against Cloudant Local.  As a workaround, you can pass the credentials in with the url (e.g. `http://user:password@host:port`).  This was discovered in #322.

emlaver:

## What

Properly handle cases when `username`, `password`, and `url` exist during client creation.  This issue was discovered when creating a client against Cloudant Local.

## How

- If `username`, `password`, and `url` exist during client creation, add `username` and `password` to `url`

## Testing
New test to assert that username and password are added to url in `reconfigure.js`.

## Issues

fixes #323

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # UNRELEASED
+- [FIXED] Case where `username` and `password` options were not used if a `url` was supplied.
 
 # 2.3.0 (2018-06-08)
 - [FIXED] Removed addition of `statusCode` to response objects returned by promises.
--- a/lib/reconfigure.js
+++ b/lib/reconfigure.js
@@ -29,6 +29,9 @@ module.exports = function(config) {
   config = JSON.parse(JSON.stringify(config)); // clone
   var outUrl;
   var outCreds = {};
+  var options;
+  var username;
+  var password;
   // if a full URL is passed in
   if (config.url) {
     // parse the URL
@@ -58,6 +61,14 @@ module.exports = function(config) {
 
       // reconstruct the URL
       config.url = url.format(parsed);
+    } else {
+      options = getOptions(config);
+      username = options.username;
+      password = options.password;
+      if (username && password) {
+        config.url = parsed.protocol + '//' + encodeURIComponent(username) + ':' +
+                    encodeURIComponent(password) + '@' + parsed.host;
+      }
     }
     outUrl = config.url;
   } else if (config.vcapServices) {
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -1,4 +1,5 @@
 # UNRELEASED
+- [FIXED] Case where if the username, password, and url existed during client creation the url was missing the required credentials.
```
ricellis: This is quite implementation specific, perhaps we can make it more user centric?
Maybe something like this:
```markdown
- [FIXED] Case where `username` and `password` options were not used if a `url` was supplied.
```
wdyt?

lib/reconfigure.js
```diff
@@ -58,6 +58,19 @@ module.exports = function(config) {
 
       // reconstruct the URL
       config.url = url.format(parsed);
+    } else {
+      var options = getOptions(config);
```
ricellis: This redefinition of variables is causing linting failures.
Probably easiest to hoist them up (as that is what happens under the covers anyway).

lib/reconfigure.js
```diff
@@ -58,6 +58,19 @@ module.exports = function(config) {
...
+      var username = options.username;
+      var password = options.password;
+
```
ricellis: Trailing space here is causing an eslint failure

test/legacy/reconfigure.js
```diff
@@ -62,6 +62,15 @@ describe('Reconfigure', function() {
...
 
+  // Issue cloudant/nodejs-cloudant#323
+  it('allows a username, a password, and url to be passed in', function(done) {
+    var credentials = { username: 'myusername', password: 'mypassword', url: 'http://localhost:8081' };
+    var outCreds = reconfigure(credentials);
+    outCreds.outUrl.should.be.a.String;
+    outCreds.outUrl.should.equal('http://myusername:mypassword@localhost:8081');
+    done();
+  });
+
```
ricellis: Perhaps another test where the `url` passed contains credentials as well?
This will validate the override behaviour, which is that `username` and `password` options supersede credentials supplied in a `url`.
I don't know if there is a reasonable place to document that as well - maybe in the `README` `Initialization>Using a URL` section?

lib/reconfigure.js
```diff
@@ -58,6 +58,19 @@ module.exports = function(config) {
 
       // reconstruct the URL
       config.url = url.format(parsed);
+    } else {
+      var options = getOptions(config);
+      var username = options.username;
+      var password = options.password;
+      
+      // remove path if it exists
+      if(parsed.host.endsWith('/')) {
```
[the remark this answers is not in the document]
emlaver: You're right, I'll remove this.

CHANGES.md
emlaver: Yea, sounds better.

emlaver:
@ricellis the `getOptions(config)` needs to be called again in the `else` block where `account` may contain `.cloudant.com` and the new value with this prefix removed is set using `if (match) { config.account = match[1]; }`.

Issue #327, opened 2018-06-27
VCAP_SERVICES support omits credentials


glynnbird:
### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue

```js
// test.js
var cloudant = require('@cloudant/cloudant');
var opts = { vcapServices: JSON.parse(process.env.VCAP_SERVICES) }

var db = cloudant(opts).db.use('animaldb')
db.list((err,data) => {
  console.log(err,data)
  // Error: _reader access is required for this request
})
```

```sh
export VCAP_SERVICES='{"cloudantNoSQLDB":[{"credentials":{"username":"X","password":"Y","host":"35d1073f-69a3-455e-aa7a-1ab424774955-bluemix.cloudant.com","port":443,"url":"https://X:Y@35d1073f-69a3-455e-aa7a-1ab424774955-bluemix.cloudant.com"},"syslog_drain_url":null,"volume_mounts":[],"label":"cloudantNoSQLDB","provider":null,"plan":"Shared","name":"iot-cloudantNoSQLDB","tags":["data_management","ibm_created","lite","ibm_dedicated_public"]}]}'
node test.js
```

### 2. What you expected to happen

I expected to get a list of documents from the database

### 3. What actually happened

- HTTP 401
- name: 'Error',
-  error: 'unauthorized',
- reason: '_reader access is required for this request',

## Environment details

- Tested on version 2.3.0
- Node.js version -  8.11.1

--------------------------

I think it stems from this section of the code https://github.com/cloudant/nodejs-cloudant/blob/master/lib/reconfigure.js#L93-L101. The `host` key is picked up from VCAP_SERVICES which is just the host name. The username and password are ignored. The `url` key has the full URL with username & password present, or the `username` & `password` can be used.

This issue was reported to me by @bradnoble who was trying to get a CloudFoundry app working with Cloudant. It worked when he supplied the URL to the library himself, but passing the parsed VCAP_SERVICES environment variable didn't work. 

I was able to reproduce this by looking at a VCAP_SERVICES env variable from one of my CloudFoundry apps and it failed in the same way for me. I verified that the credentials were missing at [this line of the library code](https://github.com/cloudant/nodejs-cloudant/blob/master/lib/reconfigure.js#L128).

------

PR #328, opened 2018-06-27
allow basic-auth url in vcap_services - fixes issue #327

glynnbird:
## Checklist

## Description

Fixed a bug in reconfigure where `vcapServices` is supplied with a basic-auth URL. 

Fixes issue #327 

## Approach

Only the `hostname` was used and the `username`/`password` was ignored. I elected to pick the `url` parameter which contains the basic-auth credentials in the absence of an `apikey` in the `credentials` object.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

 Added new tests:

- allows a vcap_services basic-auth url

## Monitoring and Logging

- "No change"

### DIFF
--- a/lib/reconfigure.js
+++ b/lib/reconfigure.js
@@ -94,6 +94,8 @@ module.exports = function(config) {
           outUrl = 'https://' + encodeURIComponent(credentials.host);
           if (credentials.apikey) {
             outCreds.iamApiKey = credentials.apikey;
+          } else if (credentials.url) {
+            outUrl = credentials.url;
           }
           break;
         } else {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # UNRELEASED
 - [FIXED] Case where `username` and `password` options were not used if a `url` was supplied.
+- [FIXED] Case where vcapServices was supplied with a basic-auth url
 
 # 2.3.0 (2018-06-08)
 - [FIXED] Removed addition of `statusCode` to response objects returned by promises.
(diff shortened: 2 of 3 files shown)


lib/reconfigure.js
```diff
@@ -94,6 +94,8 @@ module.exports = function(config) {
           outUrl = 'https://' + encodeURIComponent(credentials.host);
           if (credentials.apikey) {
             outCreds.iamApiKey = credentials.apikey;
+          } else if (credentials.url) {
```
ricellis: A "v2" `VCAP_SERVICES` `url` does not have embedded credentials, but the `host` `username` and `password` are present. I don't know if it is possible to construct a "v2" `VCAP_SERVICES` without an `apikey` - if it is then this isn't safe.
@emlaver thoughts?

emlaver: v2 VCAP, whether legacy or not, should have an `apikey`.  The non-legacy v2 VCAP has `host` and `username`, but does not have `url`  or `password`.  The legacy v2 VCAP has `host`, `username`, and `password`.
What we should do is replace the above with an `else if` block that adds username and password to url e.g.
```
...
} else if (credentials.username && credentials.password) {
  outUrl = 'https://' + encodeURIComponent(credentials.username) + ':' +
                     encodeURIComponent(credentials.password) + '@' + credentials.host;
}
```

glynnbird: ok. I'll rewrite to always construct the URL from the components. The app I was looking at *did* have a `url`, so must have been an older version of VCAP

PR #341, opened 2018-10-24
Fix plugin configuration when instantiating a new client using VCAP

### LINKED ISSUE #340: "promises" plug-in not functioning when using vcapServices option to initialize Cloudant

## Bug Description
When initializing Cloudant using the vcapServices option, the "promises" plug-in does not seem to function.  When initializing cloudant using url:<Username:Password.bluemix.cloudant.com>, the plugin DOES work.  In both cases, the Cloudant database successfully connects.  
The following code DOES work:
```javascript
let cloud_DB = new Cloudant({url:cloudand_url, plugins: 'promises' }, (err, cloudant, reply) => {
appEnv.services }, (err, cloudant, reply) => {
  if (err) {
    throw (err)
    console.log("Error creating cloudant " + err);
  }
    console.log("made DB");
});

cloud_DB.db.use('database_name').info().then((body) => {
  console.log(`got info <from a promise>: ${body}`);
})
```
the following code DOES NOT work:
```javascript
let cloud_DB = new Cloudant({vcapInstanceName:'Cloudant-d2', vcapServices: appEnv.services, plugins: 'promises' }, (err, cloudant, reply) => {
  if (err) {
    throw (err)
    console.log("Error creating cloudant " + err);
  }
    console.log("made DB");
});

cloud_DB.db.use('database_name').info().then((body) => {
  console.log(`got info <from a promise>: ${body}`);
})
```
It is fair to point out that I am passing **appEvn.services** to vcapServices.  This appEvn uses cfenv to get the vcap services or a local copy, if the code is local. 
In both cases, I would expect to see in the console:
```
made DB
got info <from a promise>: [object object]
```
### 3. What actually happened
In the first example, I see what I expect.

In the second example, I see ``made DB`` but then receive a
```TypeError: cloud_DB.db.use(...).info(...).then is not a function ```

**note** that if I make the same call without using promises, it works in both cases.  Example:
```Javascript
cloud_DB.db.use('database_name').info((err, body) => {
    if(err) console.log("DB info Error " + err);
    if(body){
      console.log("DB info " + body);
    }
});
```
## Environment details

- @cloudant/cloudant version
    > 2.4.0
- Node.js version
    > 9.8.0

smithsz:
## Checklist

## Description

Fix plugin configuration when instantiating a new client using VCAP.

Fixes #340.

## Approach

Always look at the `plugins` array for user specified plugins.

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Added additional unit test.

## Monitoring and Logging
No change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# UNRELEASED
+- [FIXED] Don't override `plugins` array when instantiating a new client using VCAP.
+
 # 2.4.0 (2018-09-19)
 - [FIXED] Case where `username` and `password` options were not used if a `url` was supplied.
 - [FIXED] Case where `vcapServices` was supplied with a basic-auth `url`.
--- a/lib/client.js
+++ b/lib/client.js
@@ -45,14 +45,19 @@ class CloudantClient {
     self._usePromises = false;
     self.useLegacyPlugin = false;
 
-    // build plugin array
+    // Build plugin array.
     var plugins = [];
-    // If IAM key found in VCAP, add IAM plugin
+
     if (self._cfg.creds && self._cfg.creds.iamApiKey) {
-      plugins.push({ iamauth: { iamApiKey: self._cfg.creds.iamApiKey } });
-    } else if (typeof this._cfg.plugin === 'undefined' && typeof this._cfg.plugins === 'undefined') {
-      plugins = [ 'cookieauth' ]; // default
-    } else {
+      // => Found IAM API key in VCAP - Add 'iamauth' plugin.
+      plugins = [ { iamauth: { iamApiKey: self._cfg.creds.iamApiKey } } ];
+    } else if (typeof this._cfg.plugins === 'undefined') {
+      // => No plugins specified - Add 'cookieauth' plugin.
+      plugins = [ 'cookieauth' ];
+    }
+
+    // Add user specified plugins.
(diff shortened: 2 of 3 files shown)


test/issues/340.js
```diff
@@ -0,0 +1,149 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* global describe it before beforeEach after */
+'use strict';
+
+const assert = require('assert');
```
ricellis: Can't we just add these new tests to the `plugins/iamauth.js` and avoid all the duplication of setup tokens etc. Having recently gone through adding the `SERVER_WITH_CREDS` stuff to allow us to test in more envs I can tell you that having this stuff in fewer places is a good thing!

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 2.5.0 UNRELEASED
```
ricellis: I think this content is just a patch release for a fix (i.e. 2.4.1). I'm happy for this to remain as `UNRELEASED` until we know the full content of the release, but I do see the value in setting it if we know that same change necessitates a particular version change.

PR #343, opened 2018-11-07
Prepare version 2.5.0 release

smithsz:
# 2.5.0 (2018-11-07)
- [FIXED] Don't override `plugins` array when instantiating a new client using VCAP.


smithsz:
See https://github.com/cloudant/nodejs-cloudant/pull/344.

PR #348, opened 2018-11-16
Prepare version 3.0.0 release

smithsz:
# 3.0.0 (2018-11-20)
- [FIXED] Expose `BasePlugin` type in Cloudant client.
- [FIXED] Set `parseUrl = false` on underlying Nano instance.
- [BREAKING CHANGE] Due to the Nano 7.x upgrade all return types are now a
  `Promise` (except for the `...AsStream` functions). _See
  [api-migration.md](https://github.com/cloudant/nodejs-cloudant/blob/master/api-migration.md#2x--3x)
  for migration details._
- [REMOVED] Remove nodejs-cloudant TypeScript type definitions for
  `db.search`. These definitions are now imported directly from Nano.
- [REMOVED] Removed support for the deprecated Cloudant virtual hosts feature.
- [UPGRADED] Using nano==7.1.1 dependency.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
-# 3.0.0 UNRELEASED
+# 3.0.0 (2018-11-20)
 - [FIXED] Expose `BasePlugin` type in Cloudant client.
 - [FIXED] Set `parseUrl = false` on underlying Nano instance.
+- [BREAKING CHANGE] Due to the Nano 7.x upgrade all return types are now a
+  `Promise` (except for the `...AsStream` functions). _See
+  [api-migration.md](https://github.com/cloudant/nodejs-cloudant/blob/master/api-migration.md#2x--3x)
+  for migration details._
 - [REMOVED] Remove nodejs-cloudant TypeScript type definitions for
   `db.search`. These definitions are now imported directly from Nano.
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,4 +1,4 @@
-# 3.0.0 UNRELEASED
+# 3.0.0 (2018-11-19)
 - [FIXED] Expose `BasePlugin` type in Cloudant client.
 - [REMOVED] Remove nodejs-cloudant TypeScript type definitions for
   `db.search`. These definitions are now imported directly from Nano.
```
ricellis: I think we need a reference to the migration guide here. Similar to what we did for [version 2](https://github.com/cloudant/nodejs-cloudant/blob/master/CHANGES.md#201-2018-02-14). I think it could be associated with the `[UPGRADED]` nano point, but perhaps we should add a `[BREAKING CHANGE]` that talks about the return type change and points to the migration guide from there e.g.

```markdown
- [BREAKING CHANGE] Due to the Nano 7.x upgrade return types are now `Promise`s. _See [api-migration.md](https://github.com/cloudant/nodejs-cloudant/blob/master/api-migration.md#2x--3x)
  for migration details._
```
wdyt?

PR #363, opened 2019-01-09
Nano8

ricellis:
## Checklist

## Description
Upgrade Apache CouchDB Nano dependency to v8 to consume removal of callbacks from `*AsStream` methods from API.

## Approach

Updated `package.json`, `package-lock.json`.

## Schema & API Changes

Nano 8 includes a breaking API change, users of callbacks with the Nano 7 API `*AsStream` methods are vulnerable to `Buffer` `RangeError` issues.

Nano 7 accepted a callback to the `*AsStream` functions, but the correct behaviour when using the `request` object from `*AsStream` functions is to use event handlers. Users of the `*AsStream` functions should ensure they are using event handlers not callbacks before moving to this version.

## Security and Privacy

- "No change"

## Testing

- Modified existing `retry-on-429 plugin #db / should return a stream` test because it used a callback with an `listAsStream`. Replaced with an event listener and corrected the mock for the endpoint (which was always incorrect, but didn't matter to the test previously).

## Monitoring and Logging

- "No change"


CHANGES.md
```diff
@@ -1,3 +1,12 @@
+# UNRELEASED
+- [BREAKING CHANGE] Nano 7 accepted a callback to the `*AsStream` functions, but
```
emlaver:
```
Users of the `*AsStream` functions should ensure they are using event handlers not callbacks before moving to this version.
```
Should we also add this as a Note to the README?

ricellis: Good point and I'll need to add something to the migration guide as well for 3 to 4.

ricellis: FYI @emlaver @smithsz - I updated the documentation in e125032 please review.

emlaver: @ricellis +1 for the doc updates

PR #386, opened 2019-08-21
Retry bad IAM token requests.

### LINKED ISSUE #385: Retry 5xx/429s token service requests in IAM plugin.
- __Token Service (https://iam.cloud.ibm.com/identity/token)__
Retry `5xx` and `429`.

- __`/_iam_session`__
Retry `5xx` (endpoint will _not_ return 429 response).

smithsz:
## Checklist

## Description
Retry bad IAM token requests.

Fixes #385.

## Approach
Catch `5xx` & `429` status codes and invoke retry mechanism (by _not_ setting `shouldApplyIAMAuth` to `false`).

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Updated unit tests.

## Monitoring and Logging
No change.

### DIFF
--- a/plugins/iamauth.js
+++ b/plugins/iamauth.js
@@ -37,9 +37,9 @@ class IAMPlugin extends BasePlugin {
 
     super(client, cfg);
 
-    this.iamApiKey = null;
+    this.currentIamApiKey = null;
     this.baseUrl = cfg.baseUrl || null;
-    this.cookieJar = request.jar();
+    this.cookieJar = null;
     this.tokenUrl = cfg.iamTokenUrl || 'https://iam.cloud.ibm.com/identity/token';
     this.shouldApplyIAMAuth = true;
     this.refreshRequired = true;
@@ -48,8 +48,10 @@ class IAMPlugin extends BasePlugin {
   onRequest(state, req, callback) {
     var self = this;
 
-    if (self._cfg.iamApiKey !== self.iamApiKey) {
-      debug('New credentials identified. Renewing session cookie...');
+    if (self._cfg.iamApiKey !== self.currentIamApiKey) {
+      debug('New IAM API key identified.');
+      this.cookieJar = request.jar(); // new jar
+      self.currentIamApiKey = self._cfg.iamApiKey;
       self.shouldApplyIAMAuth = self.refreshRequired = true;
     }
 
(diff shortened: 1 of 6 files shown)


plugins/iamauth.js
```diff
@@ -153,8 +163,10 @@ class IAMPlugin extends BasePlugin {
                 callback(new Error('Invalid response from IAM token service'));
               }
             } else {
-              self.shouldApplyIAMAuth = false;
-              callback(new Error('Failed to access token'));
+              if (response.statusCode < 500 && response.statusCode !== 429) {
+                self.shouldApplyIAMAuth = false; // no retry
```
ricellis: I think we should output a message here (or where the error callback is consumed internally) since this never bubbles up to the user. Even if we add a flag to allow it to bubble up we should still log it more visibly than debug.

plugins/iamauth.js
```diff
@@ -175,8 +187,10 @@ class IAMPlugin extends BasePlugin {
               debug('Successfully renewed IAM session.');
               callback();
             } else {
-              self.shouldApplyIAMAuth = false;
-              callback(new Error('Failed to exchange IAM token with Cloudant'));
+              if (response.statusCode < 500) {
+                self.shouldApplyIAMAuth = false; // no retry
```
ricellis: oh since we'd need to log here as well maybe logging where the callback is consumed is better

test/plugins/iamauth.js
```diff
@@ -554,4 +557,51 @@ describe('#db IAMAuth Plugin', function() {
...
+      })
+      .reply(200, MOCK_IAM_TOKEN_RESPONSE);
+
+    var cloudantMocks = nock(SERVER)
+      .post('/_iam_session', {access_token: MOCK_ACCESS_TOKEN})
+      .reply(200, {ok: true}, MOCK_SET_IAM_SESSION_HEADER)
+      .get(DBNAME)
+      .reply(200, {doc_count: 0});
+
+    var cloudantClient = new Client({ maxAttempt: 5, plugins: { iamauth: { iamApiKey: IAM_API_KEY } } });
```
ricellis: Any particular reason to test this for `5` and cause quite such a long test?
I mean we could do 1 or 2 or a shorter delay to make this a bit faster (and also potentially avoid increasing the test timeout).

plugins/iamauth.js
smithsz: Added a `debug('Disabling IAM authentication.');` in the callback.

test/plugins/iamauth.js
smithsz: Nope, will reduce to `maxAttempt: 3`.

PR #392, opened 2019-09-02
Error when user cannot be authenticated using IAM.

smithsz:
## Checklist

## Description
Error when user cannot be authenticated using IAM.

## Testing
Includes additional/updated unit tests.

## Monitoring and Logging
No change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+# UNRELEASED
+- [FIXED] Stopped disabling the IAM auth plugin after failed IAM
+  authentications. Subsequent requests will re-request authorization,
+  potentially failing again if the original authentication failure was not
+  temporary.
+- [FIXED] Ensure IAM API key can be correctly changed.
+- [FIXED] Callback with an error when a user cannot be authenticated using IAM.
+
 # 4.2.1 (2019-08-29)
 - [FIXED] Include all built-in plugin modules in webpack bundle.
 
--- a/README.md
+++ b/README.md
@@ -414,15 +414,14 @@ var cloudant = new Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: {
 
       The initial retry delay in milliseconds _(default: 500)_.
 
-   If the IAM token cannot be retrieved (either because the IAM token service is
-   down or the IAM API key is incorrect) then the request will continue without
-   IAM authentication allowing the database to return a `401` response to the
-   caller so that it may be handled appropriately.
-
    For example:
    ```js
-   var cloudant = new Cloudant({ url: 'https://examples.cloudant.com', plugins: { iamauth: { iamApiKey: 'xxxxxxxxxx', retryDelayMultiplier: 4 } } });
+   var cloudant = new Cloudant({ url: 'https://examples.cloudant.com', plugins: { iamauth: { iamApiKey: 'xxxxxxxxxx', retryDelayMultiplier: 4, retryInitialDelayMsecs: 100 } } });
    ```
+
+   If the IAM token cannot be retrieved after the configured number of retries
+   (either because the IAM token service is down or the IAM API key is
+   incorrect) then an error is returned to the client.
 
    See [IBM Cloud Identity and Access Management](https://console.bluemix.net/docs/services/Cloudant/guides/iam.html#ibm-cloud-identity-and-access-management) for more information.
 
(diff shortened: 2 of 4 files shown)


ricellis:
I don't think we need a config option here, at the end of the day if the auth exchange has failed and there are no retries remaining the request is going to fail anyway. Given the confusion we've seen around allowing the request to go and fail at the service it probably makes more sense to fail it early; making the cause more clear (i.e. the auth exchange failure).

README.md
```diff
@@ -414,16 +414,15 @@ var cloudant = new Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: {
...
-   IAM authentication allowing the database to return a `401` response to the
-   caller so that it may be handled appropriately.
-
    For example:
    ```js
-   var cloudant = new Cloudant({ url: 'https://examples.cloudant.com', plugins: { iamauth: { iamApiKey: 'xxxxxxxxxx', retryDelayMultiplier: 4 } } });
+   var cloudant = new Cloudant({ url: 'https://examples.cloudant.com', plugins: { iamauth: { iamApiKey: 'xxxxxxxxxx', retryDelayMultiplier: 4, retryInitialDelayMsecs: 100 } } });
    ```
 
+   If the IAM token cannot be retrieved (either because the IAM token service is
```
ricellis: perhaps we can augment this with "after the configured number of retries" i.e.
>If the IAM token cannot be retrieved _after the configured number of retries_

test/plugins/iamauth.js
```diff
@@ -479,39 +466,6 @@ describe('#db IAMAuth Plugin', function() {
     });
   });
 
-  it('skips authentication renewal on 401 response if previous attempts failed', function(done) {
```
ricellis: Correct me if I'm wrong - this test is no longer valid because on a `401` response from a Cloudant request we will attempt to renew the auth, but if that renewal fails the error is now returned to the user so we need no longer check that auth renewal is subsequently skipped.

test/plugins/iamauth.js
```diff
@@ -658,4 +607,56 @@ describe('#db IAMAuth Plugin', function() {
       });
     });
   });
+
+  it('returns an error if access token returns non-200 response', function(done) {
```
ricellis: It's a bit confusing that this test has the same name as the earlier test (line 345) `returns an error if access token returns non-200 response`

AFAICT the only difference is the pruned signature of the callback function in this version, perhaps that can just be applied to the previous test unless I'm missing something about the duplication.

test/plugins/iamauth.js
```diff
@@ -658,4 +607,56 @@ describe('#db IAMAuth Plugin', function() {
...
+    var cloudantClient = new Client({ plugins: { iamauth: { iamApiKey: IAM_API_KEY } } });
+    var req = { url: SERVER + DBNAME, method: 'GET' };
+    cloudantClient.request(req, function(err) {
+      assert.equal(err.message, 'Failed to acquire access token. Status code: 500');
+      iamMocks.done();
+      done();
+    });
+  });
+
+  it('returns an error if IAM cookie login returns non-200 response', function(done) {
```
ricellis: This also has the same name as the earlier test (line 403), but I can see more differences than in the previous case. Perhaps we could rename slightly if they are testing different functionality.

test/plugins/iamauth.js
smithsz: Good spot. I think this is a dupe. I'll remove it.

test/plugins/iamauth.js
smithsz: A dupe too I think. I'll remove.

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+# UNRELEASED
+- [FIXED] Always retry failed requests made by the IAM plugin.
```
ricellis: I think this could be confusing because `retry` is most often used wrt to the `retry` plugin, how about:
`Stopped disabling the IAM auth plugin after failed IAM authentications. Subsequent requests will re-request authorization, potentially failing again if the original authentication failure was not temporary.`
Bit of a mouthful 😕

CHANGES.md
```diff
@@ -1,3 +1,8 @@
...
+- [FIXED] Ensure IAM API key can be correctly changed.
+- [FIXED] Error when user cannot be authenticated using IAM.
```
ricellis: This probably needs a prefix maybe `Callback with an error...`

PR #393, opened 2019-09-17
Replace package `lockfile` with internal mutex implementation.

smithsz:
## Checklist

## Description
Replace package `lockfile` with internal mutex implementation.

## Approach
The new Mutex class provides a basic lock that can be used to prevent the simultaneous access to a resource.

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Includes additional unit testing.

## Monitoring and Logging
Adds a new log class `cloudant:mutex`.


smithsz:
This isn't quite right. It's a lot faster than lockfile but soon exceeds the stack size limit if there's a lot of concurrent lock acquire attempts. Back to the drawing board!

PR #397, opened 2019-09-23
Cookie token manager

smithsz:
## Checklist

## Description
Add `CookieTokenManager` class for managing cookie session renewals.

## Approach
The cookie renewal logic has been centralised into a token manager class. Once implemented fully, the cookie authentication plugin will make `.renew()` requests against the token manager class. The token manager class is designed to avoid unnecessary renewals and operate without needing the `lockfile` dependancy (to be removed in a later PR).

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Includes additional unit tests. Existing cookie authentication tests will be used to test `CookieTokenManager` logic.

## Monitoring and Logging
No change.

### DIFF
--- a/lib/tokens/TokenManager.js
+++ b/lib/tokens/TokenManager.js
@@ -1,0 +1,115 @@
+// Copyright © 2019 IBM Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict';
+
+const debug = require('debug')('cloudant:tokens:tokenmanager');
+const cookie = require('cookie');
+const EventEmitter = require('events');
+
+class TokenManager {
+  constructor(client, jar, sessionUrl) {
+    this._client = client;
+    this._jar = jar;
+    this._sessionUrl = sessionUrl;
+
+    this._attemptTokenRenewal = true;
+    this._isTokenRenewing = false;
+
+    this._tokenExchangeEE = new EventEmitter().setMaxListeners(Infinity);
+  }
+
+  _autoRenew() {
+    debug('Auto renewing token now...');
+    this._renew().then((response) => {
+      let maxAgeSecs = cookie.parse(response.headers['set-cookie'][0])['Max-Age'] || 3600000;
+      let delaySecs = maxAgeSecs / 2;
+      debug(`Renewing token in ${delaySecs} seconds.`);
+      setTimeout(this._autoRenew.bind(this), delaySecs * 1000).unref();
+    }).catch((error) => {
+      debug(`Failed to auto renew token - ${error}. Retrying in 60 seconds.`);
+      setTimeout(this._autoRenew.bind(this), 60000).unref();
+    });
+  }
+
+  _getToken(callback) {
+    // ** Method to be implemented by _all_ subclasses of `TokenManager` **
+    throw new Error('Not implemented.');
+  }
+
+  // Renew the token.
+  _renew() {
+    if (!this._isTokenRenewing) {
+      this._isTokenRenewing = true;
+      this._tokenExchangeEE.removeAllListeners();
+      debug('Starting token renewal.');
+      this._getToken((error, response) => {
+        if (error) {
+          this._tokenExchangeEE.emit('error', error, response);
+        } else {
+          this._tokenExchangeEE.emit('success', response);
+          this._attemptTokenRenewal = false;
+        }
+        debug('Finished token renewal.');
+        this._isTokenRenewing = false;
+      });
+    }
+    return new Promise((resolve, reject) => {
+      this._tokenExchangeEE.once('success', resolve);
+      this._tokenExchangeEE.once('error', (error, response) => {
+        error.response = response;
+        reject(error);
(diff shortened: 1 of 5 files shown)


lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,117 @@
...
+const cookie = require('cookie');
+const EventEmitter = require('events');
+
+class TokenManager {
+  constructor(client, jar, sessionUrl) {
+    this._client = client;
+    this._jar = jar;
+    this._sessionUrl = sessionUrl;
+
+    this._forceRenew = true;
```
emlaver: What about `_attemptTokenRenewal`?

lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,117 @@
...
+    this._lastRenewResponse = null;
+    this._renewingToken = false;
```
emlaver: Maybe this could be named to `_isTokenRenewing` to show that the token renewal is in progress.

lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,117 @@
...
+    // ** Method to be implemented by _all_ subclasses of `TokenManager` **
+    throw new Error('Not implemented.');
+  }
+
+  // Getter for `forceRenew`.
+  // - `true`  A renewal attempt will be made on the next `renew` request.
+  // - `false` No renewal attempt will be made on the next `renew`
+  //           request. Instead the last good renewal response will be returned
+  //           to the client.
+  get getforceRenew() {
```
emlaver: If we agree on renaming `_forceRenew` variable above, this function should also be renamed.

lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,117 @@
...
+  constructor(client, jar, sessionUrl) {
+    this._client = client;
+    this._jar = jar;
+    this._sessionUrl = sessionUrl;
+
+    this._forceRenew = true;
+    this._lastRenewResponse = null;
+    this._renewingToken = false;
+
+    this._ee = new EventEmitter().setMaxListeners(Infinity);
```
emlaver: `ee` should be updated to something more descriptive.

lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  // Renew the token if `forceRenew` is `true`. Otherwise return the last good
+  // renewal response to the client.
+  renew() {
+    if (this._forceRenew) {
+      return this.forceRenew();
+    } else {
+      return new Promise((resolve) => {
+        resolve(this._lastRenewResponse);
```
emlaver: Should we verify that `_lastRenewResponse` is not empty or null, and if it is warn or throw an error to the user?

test/tokens/TokenManager.js
```diff
@@ -0,0 +1,133 @@
...
+// limitations under the License.
+
+/* global describe it */
+'use strict';
+
+const assert = require('assert');
+
+const TokenManager = require('../../lib/tokens/TokenManager');
+
+class TestTokenManager1 extends TokenManager {
```
emlaver: Rename this so we know it's a test class containing a successfully mocked token

test/tokens/TokenManager.js
```diff
@@ -0,0 +1,133 @@
...
+    }, 100);
+  }
+
+  // mock successful token renewal
+  get getTokenCallCount() {
+    return this._getTokenCallCount;
+  }
+}
+
+class TestTokenManager2 extends TokenManager {
```
emlaver: Rename to something like `FailedTestTokenManager` to show it's using a failed mocked token

test/tokens/TokenManager.js
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  it('makes another renewal only after setting force renew', (done) => {
+    let t = new TestTokenManager1();
+    // renew 1 - make request
+    t.renew().then(() => {
+      assert.equal(t.getTokenCallCount, 1);
+      // renew 2 - return last good response
+      t.renew().then(() => {
+        assert.equal(t.getTokenCallCount, 1);
```
emlaver: Should we also assert a mocked `_lastRenewResponse`?

lib/tokens/TokenManager.js
smithsz: how about `tokenExchangeEE`?

lib/tokens/TokenManager.js
smithsz: I initially had a test looking at this response object in the resolved promise but that's gone now so I don't think we need this. If the cookie gets put in the jar and the promise is resolved then all is good. I'll remove the `_lastRenewResponse` stuff to avoid complication.

test/tokens/TokenManager.js
smithsz: See above comment. I'll remove the `_lastRenewResponse` thing.

lib/tokens/TokenManager.js
smithsz: Oh, I think the auto renew thing uses it so perhaps it can't go entirely. I'll have a ponder.

smithsz: Ner, I was wrong. It calls the internal `_renew` so we're all good!

lib/tokens/TokenManager.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  // Settter for `attemptTokenRenewal`.
+  set attemptTokenRenewal(newAttemptTokenRenewal) {
+    this._attemptTokenRenewal = newAttemptTokenRenewal;
+  }
+
+  // Renew the token if `attemptTokenRenewal` is `true`. Otherwise this is a
+  // no-op so just resolve the promise.
+  renew() {
```
ricellis: I think this should be re-named to make it more clear where it is called what is going to happen e.g.
renewIfRequired
renewIfNeeded
conditionallyRenew

or whatever

PR #395, opened 2019-09-20
Fix tests for FDB.

smithsz:
## Checklist

- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [x] Added tests for code changes _or_ test/build only changes
- [x] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [x] Completed the PR template below:

## Description
_Test only changes._

Allow the testing suite to be executed against a non-"cloudant.com" database URL (like FDB).

See FDB results [here](https://gist.github.com/smithsz/8c758280676c8c3afa646bb102e271ef). Summary of outstanding failures:

```
Cloudant
  set_security
    1) add _reader nobody role
    2) add _writer nobody role (with missing cloudant key)
    3) set couchdb_auth_only mode

Issue #292
  4) lists all query indices
  5) creates new query index

Authentication
  6) supports Authentication API - POST /_api/v2/api_keys

CORS
  7) supports CORS API - GET /_api/v2/user/config/cors
  8) supports CORS API - PUT /_api/v2/user/config/cors

Authorization
  9) supports Authorization API GET _api/v2/db/<db>/_security
  10) supports Authorization API - PUT _api/v2/db/<db/_security

Cloudant-Specific APIs
  11) Has geo querying

Cloudant Query
  12) supports Cloudant Query create indexes - POST /<db/_index API call
  13) supports Cloudant Query get indexes - GET /<db/_index
  14) supports Cloudant Query - POST /<db/_find API call
  15) supports deleting a Cloudant Query index - DELETE /db/_design/ddocname/type/viewname

Cloudant Search
  16) searches test data: author:charles
  17) searches test data: title:two

Stream
  18) bulk documents to database
  19) all documents to file
  20) all documents to file (when piping inside response handler)
```

PR #398, opened 2019-09-23
Use cookie token manager

smithsz:
## Checklist

## Description
Use the new `CookieTokenManager` (see #397) in the `cookieauth` plugin.

## Approach
Remove cookie renewal logic from plugin code and use `CookieTokenManager` instead.

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Update existing tests as follows:
- Don't include the `cookieauth` plugin by default in test clients.
- Pass the server URL to the Cloudant client (where required).
- Disable auto renewing of cookie session in tests (for easier mocking).

## Monitoring and Logging
No change.

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -26,6 +26,8 @@ const uuidv4 = require('uuid/v4'); // random
 const ME = process.env.cloudant_username || 'nodejs';
 const PASSWORD = process.env.cloudant_password || 'sjedon';
 const SERVER = process.env.SERVER_URL || `https://${ME}.cloudant.com`;
+const SERVER_NO_PROTOCOL = SERVER.replace(/^https?:\/\//, '');
+const SERVER_WITH_CREDS = `https://${ME}:${PASSWORD}@${SERVER_NO_PROTOCOL}`;
 const DBNAME = `/nodejs-cloudant-${uuidv4()}`;
 const DOCID = 'doc1';
 
@@ -88,17 +90,17 @@ describe('CloudantClient', function() {
 
   describe('plugin support', function() {
     it('get plugin path by name', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ plugins: [] });
       assert.equal(cloudantClient._buildPluginPath('dummy-plugin'), '../plugins/dummy-plugin');
     });
 
     it('get plugin path by relative path', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ plugins: [] });
       assert.equal(cloudantClient._buildPluginPath('./dummy-plugin'), path.join(process.cwd(), 'dummy-plugin'));
     });
 
     it('get plugin path by absolute path', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ plugins: [] });
       assert.equal(cloudantClient._buildPluginPath('/plugins/dummy-plugin'), '/plugins/dummy-plugin');
     });
 
@@ -111,7 +113,7 @@ describe('CloudantClient', function() {
     });
 
     it('adds cookie authentication plugin if no other plugins are specified', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ creds: { outUrl: SERVER_WITH_CREDS } });
       assert.equal(cloudantClient._plugins.length, 1);
       assert.equal(cloudantClient._plugins[0].id, 'cookieauth');
     });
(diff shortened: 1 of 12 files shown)


plugins/base.js
```diff
@@ -79,6 +80,6 @@ class BasePlugin {
 }
 
 BasePlugin.id = 'base';
-BasePlugin.pluginVersion = 2;
+BasePlugin.pluginVersion = 3;
```
smithsz: Bump plugin version due to constructor arity change.

ricellis: I think we should avoid this if possible (as it will likely mean another major).
Can't we pass the creds in the `cfg` object of the plugin constructor rather than adding a separate url arg?

ricellis: Also this file needs a copyright update

test/plugins/retry.js
```diff
@@ -33,7 +33,7 @@ describe('#db Retry Plugin', function() {
         .put(DBNAME)
         .reply(201, {ok: true});
 
-    var cloudantClient = new Client();
+    var cloudantClient = new Client({ plugins: [] });
```
ricellis: I feel like disabling plugins for a test tagged with `#db` is going to cause issues. I think perhaps this test shouldn't be tagged as such since it seems like a mock oriented thing. My original efforts at tagging were directed by the presence of the server URL, which is perhaps what led me astray here.

As you've been doing test fixes anyway - suggest you validate and remove the tag if necessary in your test fix PR.

test/readmeexamples.js
```diff
@@ -54,7 +54,7 @@ describe('#db README Examples', function() {
         .put(`/${DBNAME}`)
         .reply(201, { ok: true });
 
-      var cloudant = Cloudant({ username: ME, password: PASSWORD, url: SERVER });
+      var cloudant = Cloudant({ username: ME, password: PASSWORD, url: SERVER, plugins: [] });
```
ricellis: Similarly to above, but I think in this case the tests are required to run against the service. This is going to force them to use basic auth, is that the intent?

test/plugins/retry.js
smithsz: The default behaviour has always been to add the `cookieauth` plugin if no other plugins are specified in the client configuration. This happens in lots of places throughout the tests as we instantiate a client without configuration, i.e. `new Client()`. The problem is there are no mocks for the `POST /_session` calls that happen. This wasn't a problem before since the `cookieauth` plugin would silently swallow the assertion errors thrown by nock.

The alternative fix would be to add in the mocks. I'm happy to do this if you'd prefer. I just thought it makes the test needlessly more complicated.

ricellis: Sorry if I wasn't clear.
I'm fine with this `plugins: []` removing the `cookieauth` plugin for tests that run _only_ as mocks. Which I think is the case in `retry.js`. The aside problem is that this is tagged `#db` which effectively means run against a real DB (wrong for `retry.js` I think), but something that should be addressed in the follow-up changes to #395

test/readmeexamples.js
ricellis: I think we have to have the session mocks in this case, because the readmeexamples are definitely ones we want to run against the real service.

test/cloudant.js
```diff
@@ -77,7 +77,7 @@ describe('Cloudant #db', function() {
         .get(`/_api/v2/db/${DBNAME}/_security`)
         .reply(200, security);
 
-      var cloudant = Cloudant({ url: SERVER, username: ME, password: PASSWORD });
+      var cloudant = Cloudant({ url: SERVER, username: ME, password: PASSWORD, plugins: [] });
```
ricellis: As per my comment on the other test, disabling the cookie auth plugin forces all of our service tests to run using basic auth. I'm opposed to that given that using cookie auth is the default. I understand that the changes to the cookie model to return errors has made a change necessary here to accommodate mock testing that doesn't actually honour the cookie requests so I propose:
`plugins: []` -> `plugins: testPlugins`
where `testPlugins = (process.env.NOCK_OFF) ? [cookieauth] : []` or equivalent

wdyt?

ricellis: I think if this works as a solution we can avoid the session mocks for the readme example tests too. Sorry I didn't register earlier how pervasive the problem was and that this might be an option.

test/client.js
```diff
@@ -88,17 +90,17 @@ describe('CloudantClient', function() {
...
 
     it('get plugin path by relative path', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ plugins: [] });
       assert.equal(cloudantClient._buildPluginPath('./dummy-plugin'), path.join(process.cwd(), 'dummy-plugin'));
     });
 
     it('get plugin path by absolute path', function() {
-      var cloudantClient = new Client();
+      var cloudantClient = new Client({ plugins: [] });
```
smithsz: Client instantiation tests only.

PR #399, opened 2019-09-25
Iam token manager

smithsz:
## Checklist

## Description
Add `IAMTokenManager` class for managing cookie session renewals.

## Approach
Use token manager approach as per #397.

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing
Covered by existing unit tests.

## Monitoring and Logging
No change.

### DIFF
--- a/lib/tokens/IamTokenManager.js
+++ b/lib/tokens/IamTokenManager.js
@@ -1,0 +1,99 @@
+// Copyright © 2019 IBM Corp. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+'use strict';
+
+const a = require('async');
+const debug = require('debug')('cloudant:tokens:iamtokenmanager');
+const TokenManager = require('./TokenManager');
+
+class IAMTokenManager extends TokenManager {
+  constructor(client, jar, sessionUrl, iamTokenUrl, iamApiKey, iamClientId, iamClientSecret) {
+    super(client, jar, sessionUrl);
+
+    this._iamTokenUrl = iamTokenUrl;
+    this._iamApiKey = iamApiKey;
+    this._iamClientId = iamClientId;
+    this._iamClientSecret = iamClientSecret;
+  }
+
+  _getToken(done) {
+    var self = this;
+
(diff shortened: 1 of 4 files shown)


plugins/iamauth.js
```diff
@@ -29,165 +29,77 @@ class IAMPlugin extends BasePlugin {
...
-          } else {
-            state.retryDelayMsecs *= self._cfg.retryDelayMultiplier;
-          }
+    self._tokenManager.renew().then(() => {
+      callback(state);
+    }).catch((error) => {
+      debug(error);
+      if (state.attempt < state.maxAttempt) {
+        state.retry = true;
+        let iamResponse = error.response[0];
```
ricellis: I think I'm still mis-understanding something here, but it should be impossible to get an error from the IAM token server followed by an error from `_iam_session` because if there was an error from the token server we shouldn't even make a request to `_iam_session`.
Is the reasoning behind the array just to help you unwrap whether the error was from the first request or the second when you've lost other associated context? Can we just use a specific error type instead?

lib/tokens/TokenManager.js
```diff
@@ -32,10 +32,14 @@ class TokenManager {
...
     this._renew().then((response) => {
-      let maxAgeSecs = cookie.parse(response.headers['set-cookie'][0])['Max-Age'] || defaultMaxAgeSecs;
-      let delaySecs = maxAgeSecs / 2;
-      debug(`Renewing token in ${delaySecs} seconds.`);
-      setTimeout(this._autoRenew.bind(this), delaySecs * 1000).unref();
+      if (Array.isArray(response)) {
+        // renewal required multiple requests (i.e. IAM)
+        response = response[response.length - 1];
+      }
+      let maxAgeSecs = cookie.parse(response.headers['set-cookie'][0])['Max-Age'] || 3600000;
```
ricellis: I think likely a bad merge here, since this has removed the configurable `defaultMaxAgeSecs`

lib/tokens/IamTokenManager.js
```diff
@@ -0,0 +1,99 @@
...
+        let accessTokenAuth;
+        if (self._iamClientId && self._iamClientSecret) {
+          accessTokenAuth = { user: self._iamClientId, pass: self._iamClientSecret };
+        }
+        debug('Getting access token.');
+        self._client({
+          url: self._iamTokenUrl,
+          method: 'POST',
+          auth: accessTokenAuth,
+          headers: { 'Accepts': 'application/json' },
```
ricellis: Does this need to specify the `Content-Type` of `application/x-www-form-urlencoded` as well?

plugins/iamauth.js
ricellis: @smithsz any thoughts on this

smithsz: As discussed, I'll return only the last response in the array to avoid any confusion.

lib/tokens/IamTokenManager.js
smithsz: Nope. Request does this for us.

PR #405, opened 2019-10-15
Disable cookie authentication if credentials are missing.

smithsz:
## Checklist

## Description
Disable cookie authentication if credentials are missing.

## Approach
If credentials are missing then set a 'disabled' property on the cookie plugin to prevent hooks from being executed.

## Schema & API Changes
No change.

## Security and Privacy
No change.

## Testing

## Monitoring and Logging
No change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 - [FIXED] Ensure IAM API key can be correctly changed.
 - [FIXED] Callback with an error when a user cannot be authenticated using IAM.
 - [FIXED] Ensure authorization tokens are not unnecessarily requested.
+- [IMPROVED] Do not apply cookie authentication by default in the case that no
+  credentials are provided.
 - [IMPROVED] Preemptively renew authorization tokens that are due to expire.
 
 # 4.2.1 (2019-08-29)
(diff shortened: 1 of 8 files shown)


plugins/cookieauth.js
```diff
@@ -32,7 +32,9 @@ class CookiePlugin extends BasePlugin {
     let sessionUrl = new u.URL(cfg.serverUrl);
     sessionUrl.pathname = '/_session';
     if (!sessionUrl.username || !sessionUrl.password) {
```
ricellis: I think we could add another flag to this check e.g. `errorOnNoCreds` so that when the cookie plugin is explicitly added by a user, but has no credentials that flag would default to `true` and they would receive an error for the misconfiguration. When we add the plugin by default we can set it to `false` in the configuration and allow the no auth case to not error.

CHANGES.md
```diff
@@ -6,6 +6,8 @@
 - [FIXED] Ensure IAM API key can be correctly changed.
 - [FIXED] Callback with an error when a user cannot be authenticated using IAM.
 - [FIXED] Ensure authorization tokens are not unnecessarily requested.
+- [IMPROVED] Disable cookie authentication if credentials are missing and the
```
ricellis: How about `Do not apply cookie authentication by default in the case that no credentials are provided.`

PR #411, opened 2019-11-29
Improved docs for initialization callback

### LINKED ISSUE #409: 401 from Cloudant when require_valid_user=true

## Bug Description
Two related problems. 

Switching on `require_valid_user=true` has stopped our clients connecting to Cloudant via the client library in this repo. The `POST _session` call always returns a 401, despite the correct username and password being provided (in the url or as separate params). A CURL equivalent works ok and returns 200. Here is the code for our call.

`that.cloudantInit({ url: CDB_URL, plugins: cloudantRequestRetry }, function (err, cloudant) {...})`

Given that CURL works, it must be something the client is doing. Note that we do not provide `cookieauth` as a plugin. However, when we switched on client debug, we could see that the `cookieauth` plugin was **always** being added, whatever value we specified for the `plugins` parameter (a function, empty array, omitted). It turns out that the following line of code is always executed, and is what is adding `cookieauth`...

https://github.com/cloudant/nodejs-cloudant/blob/master/cloudant.js#L300

This is in contradiction to the readme. When that line is disabled, clients connect ok with 200. But are unable to take advantage of session cookies.

So two problems:
1) `cookieauth` is always added, no way to stop that.
2) When `cookieauth` is added, clients get 401 when `require_valid_user=true`.

Catch-22.

We are approaching a code ship deadline, so urgently need a resolution to 1) and preferably 2) as well.

Set `require_valid_user=true` in Cloudant `.ini` file, start Cloudant, connect using the client, do not specify `cookieauth` plugin.

The client should connect with 200. 

### 3. What actually happened
Client gets 401.

## Environment details
- Version(s) that are affected by this issue.
    > 4.2.2
- Node.js version
    > 10.15

ricellis:
## Checklist

## Description

Improve documentation around initialization callback.
Fixes #409

## Approach

Make it clear in the documentation that using the intialization callback will always enable the `cookieauth` plugin.
Provide an example of using the `ping` function as an alternative to the initialization callback

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- N/A doc only changes

## Monitoring and Logging

- "No change"

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ Note, if you only have a single Cloudant service then specifying the
 
 You can optionally provide a callback to the Cloudant initialization
 function. This will make the library automatically
-[ping](https://console.bluemix.net/docs/services/Cloudant/api/account.html#ping)
+[ping](https://cloud.ibm.com/docs/services/Cloudant/api?topic=cloudant-account#ping)
 Cloudant to confirm the connection and that your credentials work.
 
 Here is a simple example of initializing asynchronously, using its optional
@@ -253,7 +253,7 @@ Cloudant({ account: me, password: password }, function(err, cloudant, pong) {
 initialization callback will always add the `cookieauth` plugin to the list of
 configured plugins. If you do not want to use `cookieauth` then you should not
 use this initialization callback, instead initalize the client without a
-callback and then call the [ping function](https://console.bluemix.net/docs/services/Cloudant/api/account.html#ping)
+callback and then call the [ping function](https://cloud.ibm.com/docs/services/Cloudant/api?topic=cloudant-account#ping)
 e.g.:
 
 ```js
@@ -406,7 +406,7 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
 
    IBM Cloud Identity & Access Management enables you to securely authenticate
    users and control access to all cloud resources consistently in the IBM
-   Bluemix Cloud Platform.
+   Cloud Platform.
 
    This plugin will automatically exchange your IAM API key for a token. It will
    handle the authentication and ensure that the token is refreshed as required.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -249,6 +249,25 @@ Cloudant({ account: me, password: password }, function(err, cloudant, pong) {
...
+
+```js
+const c = new Cloudant({
+  url: 'http://localhost:5984',
+  username: 'somebody',
+  password: 'something',
+  plugins: [] // disable cookieauth
+});
+c.ping()
+  .then(resp => /*do something with response*/ console.log(resp));
```
glynnbird: I don't think you want a semi-colon at the end of this line

glynnbird: Did you know you can do:

```
c.ping().then(console.log).catch(console.error)
```

just to make it less understandable ;) ?

README.md
```diff
@@ -249,6 +249,25 @@ Cloudant({ account: me, password: password }, function(err, cloudant, pong) {
 });
 ~~~
 
+**Note:** For legacy compatibility with older versions of the library using the
+initialization callback will always add the `cookieauth` plugin to the list of
+configured plugins. If you do not want to use `cookieauth` then you should not
+use this initialization callback, instead initalize the client without a
+callback and then call the [ping function](https://console.bluemix.net/docs/services/Cloudant/api/account.html#ping)
```
glynnbird: ping function is now here: https://cloud.ibm.com/docs/services/Cloudant/api?topic=cloudant-account#ping

README.md
ricellis: Oof, sloppy.
And neato.
Fixed in 50a96eadab3837d07dfa9aa8dda300a6b5c13006

README.md
ricellis: Nice spot, looks like we had a lot of these around still.
Fixed in 1b10733822c893642bbe5cee9d773b623b1742f5

PR #412, opened 2019-12-03
Used application/json for session POST

ricellis:
## Checklist

## Description

Use `application/json` for `_session` POST of cookie auth

Pass a `url` containing percent-encoded user information.
cookieauth plugin session authentication to work correctly
### 3. What actually happened
`401 unauthorized` because the credentials were double encoded.

## Approach

* Decoded credentials from URL before passing to `CookieTokenManager`.
    * Prevents double encoding if passed in a form body
* Modified CookieTokenManager to use `json` body instead of `form` encoded anyway
* Improved documentation of characters that must be encoded.

## Schema & API Changes

- No change

## Security and Privacy

Insignificant changes to handling of credentials
* For cookieauth credentials are now posted in a JSON body instead of as form encoded data, but it is the same request body as previously (i.e. if the server is https then the body containing the creds will be encrypted)
* Preferentially credentials should be passed in the `username` and `password` configuration options instead, but in the case where they are passed in the URL user info it is now documented what characters must be encoded.

## Testing

- Modified existing `cookieauth` tests to use a password with special characters.
- Additional manual testing of credentials passed in URL.

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 - [FIXED] Expose BasePlugin.
+- [FIXED] Prevent double encoding of credentials passed in URL user information
+  when using the `cookieauth` plugin.
+- [IMPROVED] Documented the characters that are required to be encoded in URL
+  user information.
 - [IMPROVED] Documented the legacy compatibility behaviour that always adds the
   `cookieauth` plugin when using the initialization callback functionality.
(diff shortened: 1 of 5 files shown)


ricellis:
@bessbd I tried to make the doc even more clear with an example in 245cbb0

PR #419, opened 2020-03-23
Token refreshment memory-leak

vmatyusGitHub:
## Checklist

## Description

There is a memory leak when the client is instantiated inside a http request.
``` javascript
var Cloudant = require('./cloudant');
const http = require('http');

const server = http.createServer((req, res) => {
  var cloudantBasic = Cloudant({
    //account setting
    plugins: 'cookieauth'
  });

  async function getDbList() {
    return cloudantBasic.db.list();
  }

  getDbList().then(result => {
    res.statusCode = 200;
    res.end(JSON.stringify(result) + '\n');
    console.log('####test success');
  }).catch(err => {
    res.statusCode = 401;
    res.end(err.error + '\n');
    console.log('####test failed', err);
  });
});

server.listen(3000);
console.log('Server listening to port 3000. Press Ctrl+C to stop it.');
```
In this case when the server receives a request it will instantiate a client and after the response arrived all the request that went out toward Cloudant remained in the memory.

## Approach

The default `CloudantClient` has the `keepAlive` option turned on and thank to this the opened sockets remain open and the last sent requests remain in the memory.
So the memory could have been cleared when every socket had been closed. 
The current library supports only singleton usage. If it is not followed by the user there will be memory leak in the application.
However, to mitigate the problem I used `bind` functions to remove several references from the memory and created a specific Agent for token refreshment, where the `keepAlive` is set to `false`. This way the token refreshment requests will be closed when the request is completed.

Mainly modified components: TokenManager, CoockieTokenManager, IAMTokenManager.

The above code sample was executed with the previous library and the current version. I sent 8 request to see the memory consumption:
|---| prev | now | diff |
|---|---|---|---|
| Request speed | 7087 ms | 7232 ms | +200ms |
| after 8 request | 8.8 MB | 8.9 MB | +0.1MB |
| 5 minutes later | 8.1 MB | 7.8 MB | -0.3MB |

After 5 minutes more memory could be freed up by the Garbage Collector.

## Schema & API Changes
Not changed.

## Security and Privacy
Not changed.

## Testing
Updated 1 test case:
 - `makes another renewal only after setting force renew`: added `done` closure to `catch` method of the Promise. This way the test case will not remain in hanging if it fails.

## Monitoring and Logging

Modified `Finished token renewal.` to `Answer received to token renewal request.`
Added 2 new debug log:
- `Succeeded to renew token`
- `Failed to renew token`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 # 4.2.5 (2020-03-30)
-- [FIXED] Reduce the memory usage of automatical token/cookie refreshment
+- [IMPROVED] Reduce the memory usage of automatical token/cookie refreshment
 
 # 4.2.4 (2020-03-02)
 - [FIXED] Pinned Nano to version 8.1 to resolve issue with extending upstream
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 # 4.2.5 (2020-03-30)
-- [IMPROVED] Reduce the memory usage of automatical token/cookie refreshment
+- [IMPROVED] Reduce the memory usage of automatic token/cookie refreshment.
 
 # 4.2.4 (2020-03-02)
 - [FIXED] Pinned Nano to version 8.1 to resolve issue with extending upstream
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# 4.2.5 (2020-03-30)
+- [FIXED] Reduce the memory usage of automatical token/cookie refreshment
+
 # 4.2.4 (2020-03-02)
 - [FIXED] Pinned Nano to version 8.1 to resolve issue with extending upstream
   TypeScript changes in Nano version 8.2.0.
--- a/README.md
+++ b/README.md
@@ -145,9 +145,9 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result
 of calling `Cloudant()`).
 
-The \`cloudant\` client connection utilizes an agent with a configurable [HTTP
+The `cloudant` client connection utilizes an agent with a configurable [HTTP
 connection pool](#pool-size-and-open-sockets). As such the performance of the
-client is improved when re-using the \`cloudant\` client connection throughout
+client is improved when re-using the `cloudant` client connection throughout
 an application instead of repeatedly re-instantiating. It is important to
 instantiate the `cloudant` _connection_ **only once** during the application
 lifetime to reduce the overheads of memory usage and other resources such as
--- a/README.md
+++ b/README.md
@@ -145,6 +145,11 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result of
 calling `Cloudant()`).
 
+If auto token refreshment is used (as default it is activated, see more details
+[here](#plugin-configuration)) then it is important to define one connection **only
+once** during the application lifetime. The package do not supply connection
+shutdown functinality.
+
 You can initialize your client in _one_ of the following ways:
 
 #### 1. Using a URL:
@@ -396,7 +401,9 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
 
 1. `cookieauth`
 
-   This plugin will automatically exchange your Cloudant credentials for a
+   If there is no plugin specified this will be the default plugin.
+
+   It will automatically exchange your Cloudant credentials for a
    cookie. It will handle the authentication and ensure that the cookie is
    refreshed as required.
 
@@ -414,6 +421,11 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
 
    If you don't specify a username and password during the client construction
    then cookie authentication is disabled.
+
+   You can turn off automatically refreshing cookie with the following configuration:
+   ```js
+   var cloudant = Cloudant({ url: 'https://user:pass@examples.cloudant.com', plugins: [ { cookieauth: { autoRenew: false } } ] });
+   ```
 
 2. `iamauth`
 
@@ -445,6 +457,11 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
    incorrect) then an error is returned to the client.
 
    See [IBM Cloud Identity and Access Management](https://cloud.ibm.com/docs/services/Cloudant/guides?topic=cloudant-ibm-cloud-identity-and-access-management-iam-) for more information.
+
+   You can turn off automatically refreshing token with the following configuration:
+   ```js
+   var cloudant = Cloudant({ url: 'https://user:pass@examples.cloudant.com', plugins: [ { iamauth: { iamApiKey: 'xxxxxxxxxx', autoRenew: false } } ] });
+   ```
 
 3. `retry`
 
--- a/README.md
+++ b/README.md
@@ -142,11 +142,16 @@ promises and callback styles. The library supports them all.
 
 To use Cloudant, add `require('@cloudant/cloudant')` in your code. The common
 style is that `Cloudant` (upper-case) is the **package** you load; whereas
-`cloudant` (lower-case) is your **connection** to your database (i.e. the result of
-calling `Cloudant()`).
+`cloudant` (lower-case) is your **connection** to your database (i.e. the result
+of calling `Cloudant()`).
 
- It is important to define one _connection_ **only once** during the application lifetime.
- The package do not support closing server connection.
+The \`cloudant\` client connection utilizes an agent with a configurable [HTTP
+connection pool](#pool-size-and-open-sockets). As such the performance of the
+client is improved when re-using the \`cloudant\` client connection throughout
+an application instead of repeatedly re-instantiating. It is important to
+instantiate the `cloudant` _connection_ **only once** during the application
+lifetime to reduce the overheads of memory usage and other resources such as
+unused connections.
 
 You can initialize your client in _one_ of the following ways:
 
--- a/README.md
+++ b/README.md
@@ -145,10 +145,8 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result of
 calling `Cloudant()`).
 
-If auto token refreshment is used (as default it is activated, see more details
-[here](#plugin-configuration)) then it is important to define one connection **only
-once** during the application lifetime. The package do not supply connection
-shutdown functinality.
+ It is important to define one _connection_ **only once** during the application lifetime.
+ The package do not support closing server connection.
 
 You can initialize your client in _one_ of the following ways:
 
@@ -403,7 +401,7 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
 
    If there is no plugin specified this will be the default plugin.
 
-   It will automatically exchange your Cloudant credentials for a
+   This plugin will automatically exchange your Cloudant credentials for a
    cookie. It will handle the authentication and ensure that the cookie is
    refreshed as required.
 
--- a/cloudant.js
+++ b/cloudant.js
@@ -290,12 +290,7 @@ function Cloudant(options, callback) {
       body: configuration }, callback);
   };
 
-  var close = function() {
-    nano.cc.stopTokenRefresh();
-  };
-
   // add top-level Cloudant-specific functions
-  nano.close = close;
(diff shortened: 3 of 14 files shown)


lib/tokens/IamTokenManager.js
```diff
@@ -90,6 +90,13 @@ class IAMTokenManager extends TokenManager {
     ], (error, responses) => {
       done(error, responses[responses.length - 1]);
     });
+    setTimeout(function() {
+      debug('Destroying IAM token requests');
+      reqToken.abort();
```
smithsz: Can you add a comment here please explaining why this is necessary (referencing the memory leak etc)?

lib/tokens/CookieTokenManager.js
```diff
@@ -35,19 +38,26 @@ class CookieTokenManager extends TokenManager {
...
-        let msg = `Failed to get cookie. Status code: ${response.statusCode}`;
-        debug(msg);
-        callback(new Error(msg), response);
-      }
-    });
+    }, boundedCompleted);
+
+    setTimeout(function() {
+      debug('Destroying cookie token request');
+      req.abort();
```
smithsz: And a comment here too please...

cloudant.js
```diff
@@ -290,7 +290,12 @@ function Cloudant(options, callback) {
       body: configuration }, callback);
   };
 
+  var close = function() {
+    nano.cc.stopTokenRefresh();
+  };
+
   // add top-level Cloudant-specific functions
+  nano.close = close;
```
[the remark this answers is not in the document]
vmatyusGitHub: So regarding the content discussion. The client closing functionality was removed form the code base.

lib/tokens/IamTokenManager.js
vmatyusGitHub: Instead of this approach I created a new Agent that has no `keepAlive` set to `true` This way every token refreshment request will be removed from the memory when they completed.

lib/tokens/CookieTokenManager.js
```diff
@@ -26,28 +26,48 @@ class CookieTokenManager extends TokenManager {
...
+
+    let protocol;
+    if (this._sessionUrl.startsWith('https')) {
+      protocol = require('https');
+    } else {
+      protocol = require('http');
+    }
+    let tokenAgent = new protocol.Agent({
+      keepAlive: false,
+      maxSockets: 3
```
ricellis: This agent is only used once at any given time for a cookie request, I think we may as well use `maxSockets: 1` wdyt?

lib/tokens/IamTokenManager.js
```diff
@@ -32,6 +32,16 @@ class IAMTokenManager extends TokenManager {
 
     debug('Making IAM session request.');
     let accessToken;
+    let protocol;
```
ricellis: this agent creation block is pretty much identical to that in the `CookieTokenManager` - I suggest pulling the agent code up into the `TokenManager` so that it can be written/maintained in one place and used by both `CookieTokenManager` and `IamTokenManager`.

lib/tokens/CookieTokenManager.js
```diff
@@ -26,28 +26,48 @@ class CookieTokenManager extends TokenManager {
...
   _getToken(callback) {
     debug('Submitting cookie token request.');
+    let boundedCompleted = this._reqCompleted.bind({
+      cb: callback
+    });
+
+    let protocol;
+    if (this._sessionUrl.startsWith('https')) {
+      protocol = require('https');
+    } else {
```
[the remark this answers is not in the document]
vmatyusGitHub: Not really. There are protocol checks at these parts of the code:
 - https://github.com/cloudant/nodejs-cloudant/blob/35dbb085857bb423bc86aef10c4e27a302e1f113/cloudant.js#L63-L65
 - https://github.com/cloudant/nodejs-cloudant/blob/35dbb085857bb423bc86aef10c4e27a302e1f113/lib/client.js#L188-L193

I tried to use the same logic. But I can replace the url check with this solution (just to be consistent):
https://github.com/cloudant/nodejs-cloudant/blob/35dbb085857bb423bc86aef10c4e27a302e1f113/cloudant.js#L63

package.json
```diff
@@ -46,13 +45,16 @@
...
     "mocha": "^5.1.1",
-    "nock": "^9.2.5",
+    "nock": "^12.0.3",
     "should": "6.0.3",
+    "tslint": "^5",
     "typescript": "^2.8.3",
     "uuid": "^3.0.1"
   },
   "scripts": {
+    "preinstall": "npx npm-force-resolutions",
```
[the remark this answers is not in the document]
vmatyusGitHub: There is a security vulnerability problem detected in the `minimist`, so `npm audit` command drops a warning if it detects wrong `minimist` version. The automated testing pipeline calls it here:
https://github.com/cloudant/nodejs-cloudant/blob/35dbb085857bb423bc86aef10c4e27a302e1f113/package.json#L56

The problem is that `eslint` and `mocha` packages are using this not secure version. This way the testing pipeline stops at the audit check.
With this `preinstall` command I could set to use a preinstalled `minimist` version.
I used this article as a solution base: https://itnext.io/fixing-security-vulnerabilities-in-npm-dependencies-in-less-than-3-mins-a53af735261d
Before merge I thought to check if there are any corrected package available from them.

README.md
```diff
@@ -145,6 +145,9 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result of
 calling `Cloudant()`).
 
+ It is important to define one _connection_ **only once** during the application lifetime.
+ The package do not support closing server connection.
```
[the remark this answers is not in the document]
vmatyusGitHub: This comment will overwrite this part: https://github.com/cloudant/nodejs-cloudant/pull/419#discussion_r401026960

package.json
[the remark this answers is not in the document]
mojito317: I could only guess that you changed it because of the pipeline. Thanks for the thorough explanation! You have spent way more time on this solution to not spend more, this solution is fine for me.

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 4.2.5 (2020-03-30)
+- [IMPROVED] Reduce the memory usage of automatical token/cookie refreshment
```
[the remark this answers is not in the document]
vmatyusGitHub: Changed in 00eccec

README.md
```diff
@@ -145,6 +145,9 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result of
 calling `Cloudant()`).
 
+ It is important to define one _connection_ **only once** during the application lifetime.
```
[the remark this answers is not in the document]
vmatyusGitHub: Changed in 00eccec

lib/tokens/CookieTokenManager.js
vmatyusGitHub: Changed in 00eccec

Yes, correct. I couldn't dare to set it to 1, because it is always good if there is a backup(+1). But in this case, it is not valid. Thanks for the encouragement.

lib/tokens/CookieTokenManager.js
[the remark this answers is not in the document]
mojito317: Since a possible exception is not handled at the entry point of the client I guess it is not your task to handle it here. I am ok with it if you leave this as-is.

README.md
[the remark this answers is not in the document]
ricellis: I had tried to escape formatting to make the new content easier to copy/paste, but it looks like the escaping backticks made it through into the commit 00eccec - that will need correcting so that \`cloudant\` becomes correctly marked up as `cloudant`.
Sorry for the confusion there.

lib/tokens/CookieTokenManager.js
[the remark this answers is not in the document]
vmatyusGitHub: I changed in 89d3fb0eb29f3b36000ddc34844c3eb8757cf640

README.md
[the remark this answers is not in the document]
vmatyusGitHub: Corrected in 89d3fb0eb29f3b36000ddc34844c3eb8757cf640

lib/tokens/IamTokenManager.js
vmatyusGitHub: During the refactoring I was also removed the agent reference from the Token Manager, because the new design does not need it anymore.
Thanks to this change the memory usage got cleaner surprisingly. I will soon update the PR description with the results.
The changes can be found here: 89d3fb0eb29f3b36000ddc34844c3eb8757cf640
Please let me know if there will be needed to make additional tests.

lib/client.js
```diff
@@ -132,7 +132,7 @@ class CloudantClient {
         var creds = self._cfg.creds || {};
         self._plugins.push(
           // instantiate plugin
-          new Plugin(self._client, Object.assign({ serverUrl: creds.outUrl }, cfg))
+          new Plugin(self._client, Object.assign({ serverUrl: creds.outUrl }, cfg), self.requestDefaults)
```
ricellis, line 135: Do we really need to change this constructor? IIUC `self._client` is a `request` object already and we already pass that so we can just call `.defaults()` on that object in the plugins.

lib/tokens/TokenManager.js
```diff
@@ -55,7 +55,28 @@ class TokenManager {
...
+        protocol = require('http');
+      } else {
+        protocol = require('https'); // default to https
+      }
+      var tokenAgent = new protocol.Agent({
+        keepAlive: false,
+        maxSockets: 1
+      });
+      let defaults = {};
+      defaults = Object.assign(defaults, this._reqDefault);
```
ricellis, line 70: I think this line is unnecessary.

lib/tokens/TokenManager.js
```diff
@@ -55,7 +55,28 @@ class TokenManager {
...
+      defaults.agent = tokenAgent;
+      if (defaults.headers === undefined) {
+        // if there are custom client used there will be no default request settings.
+        defaults.headers = {'Connection': 'closed'};
+      } else {
+        defaults.headers['Connection'] = 'closed';
+      }
+      let tokenRequest = require('request').defaults(defaults);
```
ricellis, line 78: I think this can be:
`_client.defaults(defaults)`

plugins/cookieauth.js
```diff
@@ -24,7 +24,7 @@ const CookieTokenManager = require('../lib/tokens/CookieTokenManager');
  * Cookie Authentication plugin.
  */
 class CookiePlugin extends BasePlugin {
-  constructor(client, cfg) {
+  constructor(client, cfg, reqDefaults) {
```
ricellis, line 27: I think this constructor can remain the same and we do either:
client.defaults() to get the requestDefaults to pass through to the TokenManager or just pass the client through, and call `defaults()` on it in the  TokenManager` - assuming that isn't the cause of the memory issue.

plugins/iamauth.js
```diff
@@ -24,7 +24,7 @@ const IAMTokenManager = require('../lib/tokens/IamTokenManager');
  * IAM Authentication plugin.
  */
 class IAMPlugin extends BasePlugin {
-  constructor(client, cfg) {
+  constructor(client, cfg, reqDefaults) {
```
ricellis, line 27: Same as for `CookiePlugin`

lib/tokens/CookieTokenManager.js
```diff
@@ -17,37 +17,22 @@ const debug = require('debug')('cloudant:tokens:cookietokenmanager');
 const TokenManager = require('./TokenManager');
 
 class CookieTokenManager extends TokenManager {
-  constructor(client, jar, sessionUrl, username, password) {
-    super(client, jar, sessionUrl);
+  constructor(rDefaults, jar, sessionUrl, username, password) {
```
ricellis, line 20: I think this can stay as it was with the client.

lib/tokens/IamTokenManager.js
```diff
@@ -18,45 +18,33 @@ const debug = require('debug')('cloudant:tokens:iamtokenmanager');
 const TokenManager = require('./TokenManager');
 
 class IAMTokenManager extends TokenManager {
-  constructor(client, jar, sessionUrl, iamTokenUrl, iamApiKey, iamClientId, iamClientSecret) {
-    super(client, jar, sessionUrl);
+  constructor(rDefaults, jar, sessionUrl, iamTokenUrl, iamApiKey, iamClientId, iamClientSecret) {
```
ricellis, line 21: Same as for `CookieTokenManager`

lib/tokens/IamTokenManager.js
vmatyusGitHub: What do you think about this aspect: a77b91f667fd

vmatyusGitHub:
A suggestion of usage about Cloudant client was added in this https://github.com/cloudant/nodejs-cloudant/pull/420

PR #420, opened 2020-04-28
Memory leak readme

vmatyusGitHub:
## Checklist

## Description

Under some circumstances there is a memory leak when the Cloudant client is instantiated more than one times during an application lifetime.
``` javascript
var Cloudant = require('./cloudant');
const http = require('http');

const server = http.createServer((req, res) => {
  var cloudantBasic = Cloudant({
    //account setting
    plugins: 'cookieauth'
  });

  async function getDbList() {
    return cloudantBasic.db.list();
  }

  getDbList().then(result => {
    res.statusCode = 200;
    res.end(JSON.stringify(result) + '\n');
    console.log('####test success');
  }).catch(err => {
    res.statusCode = 401;
    res.end(err.error + '\n');
    console.log('####test failed', err);
  });
});

server.listen(3000);
console.log('Server listening to port 3000. Press Ctrl+C to stop it.');
```
In this case when the server receives a request it will instantiate a client and after it is answered all the request that went out toward Cloudant will remain in the memory.

## Approach

During the investigation I used Chrome inspector tool, package `heapdump` to create a snapshot about the memory usage and `process.memoryUsage` to see the size of the memory reservations.
As the result of the investigation it was determined the default `Agent` keeps the network connections open for further usage thanks to setting `keepAlive=false`. This settings improves the response times and has performance advantages. This way it will be kept. 
We agreed not to change the agent storing, but extend the README.md with the supported client usage best practice. This way code was not change.

## Schema & API Changes
No change

## Security and Privacy
Correct vulnerability issue with package `minimist`:
https://www.npmjs.com/advisories/1179

## Testing
I created a script that measured the memory usage in of both of the token handling instances: Cookie and IAM. This script is not committed.
No new test case added.

## Monitoring and Logging
No change

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "eslint-plugin-react": "^7.0.0",
     "eslint-plugin-standard": "^3.0.1",
     "md5-file": "^3.2.3",
-    "mocha": "^6.2.3",
+    "mocha": "^7.1.2",
     "nock": "^9.2.5",
     "should": "6.0.3",
     "typescript": "^2.8.3",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "eslint-plugin-react": "^7.0.0",
     "eslint-plugin-standard": "^3.0.1",
     "md5-file": "^3.2.3",
-    "mocha": "^5.1.1",
+    "mocha": "^6.2.3",
     "nock": "^9.2.5",
     "should": "6.0.3",
     "typescript": "^2.8.3",
--- a/README.md
+++ b/README.md
@@ -145,6 +145,14 @@ style is that `Cloudant` (upper-case) is the **package** you load; whereas
 `cloudant` (lower-case) is your **connection** to your database (i.e. the result of
 calling `Cloudant()`).
 
+The `cloudant` client connection utilizes an agent with a configurable [HTTP
+connection pool](#pool-size-and-open-sockets). As such the performance of the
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "eslint-plugin-react": "^7.0.0",
     "eslint-plugin-standard": "^3.0.1",
     "md5-file": "^3.2.3",
-    "mocha": "^5.1.1",
+    "mocha": "^6.2.3",
```
ricellis, line 49: Does this mean that the `minimist` vulnerability warning was resolved upstream? i.e. we don't need to force a resolution any more?

vmatyusGitHub: Yes, exactly. This new mocha version was released recently contains the correction of the vulnerability problem: https://github.com/mochajs/mocha/releases/tag/v6.2.3. 
Furthermore, the pipeline is green. If there would have been some `audit` problem, then it would be red.

vmatyusGitHub: An alternative mocha version is also available with the correction: https://github.com/mochajs/mocha/releases/tag/v7.1.2
It contains the latest `mkdirp` version: 0.5.5 This `mkdirp` package uses `minimist` package. `mkdirp` v0.5.4  is the first version where the corrected `minimist` in use and this version is used in mocha v6.2.3

ricellis: OK well we could just update to 7.1.2 then if the `6`->`7` doesn't cause any failures.

PR #423, opened 2020-06-16
Concatenate streaming data in client tests

emlaver:
## Checklist

## Description
Some of the tests assert on streaming response data in the `data` callback.  This can cause a failure since the assertion will fail in a case where there's more than one chunk of streaming data.  This PR concatenates the streaming data then asserts once the stream is completed in `end` block.

## Approach
- Create a `responseString` variable to hold the streaming data and then assert on this variable once the stream is complete
- Removed failing and redundant data chunk count assertion.  We are already asserting on the streaming data

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -955,7 +955,7 @@ describe('CloudantClient', function() {
           method: 'GET'
         };
 
-        var responseString = '';
+        var responseData = '';
         cloudantClient.request(options)
           .on('error', function(err) {
             assert.fail(`Unexpected error: ${err}`);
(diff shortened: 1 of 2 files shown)


test/client.js
```diff
@@ -954,6 +954,8 @@ describe('CloudantClient', function() {
           auth: { username: ME, password: PASSWORD },
           method: 'GET'
         };
+
+        var responseString = '';
```
ricellis, line 958: it's a bit weird to call this responseString and then have to `toString` it later.
I think `responseData` or `responseBytes` probably makes more sense.

test/plugins/retry.js
```diff
@@ -289,12 +290,11 @@ describe('Retry Plugin', function() {
             assert.equal(resp.statusCode, 200);
           })
           .on('data', function(data) {
-            dataCount++;
-            assert.ok(data.toString('utf8').indexOf('"doc_count":0') > -1);
+            responseString += data;
           })
           .on('end', function() {
+            assert.ok(responseString.toString('utf8').indexOf('"doc_count":0') > -1);
```
ricellis, line 296: Do we even need to collect the chunks and assert the string here and then assert the object again later in the `on('finish'...`?
Hmm actually I suppose it does mean that we can distinguish between a failure of "expected `data` didn't arrive" vs a failure of the "plugin didn't work correctly with `pipe`".

PR #424, opened 2020-07-15
Add cache support for IAM token

hermansb:
## Checklist

## Description

The problem this PR is trying to address is the issue of excessive IAM API key exchanges for access tokens. We share the IAM API key to interact with our cloudant account with all squads in our tribe and we were hitting the rate limit that IAM had set for API key exchanges. Some apps were exchanging tokens more than others, for example our loopback v3 app which exchanged tokens for each cloudant database configured with loopback however there was room for improvement in all apps as each kubernetes pod would request its own token. Approach used in this PR is below

## Approach

In order to prevent unnecessary IAM api token exchanges here we provide support for a cache object requiring two methods: get(key) and set(key, value, ttl) where ttl is the time to live. We also provide support for another parameter `cacheOffsetInSecs` which subtracts from the ttl to store in the cache so that we don't cut it too close to the expiry time. The cache should not return the token if the token is expired and instead allow the library to retrieve a new access token from IAM. The get and set API methods should return promises.

## Schema & API Changes

Two new config parameters are now supported in the `iamauth` plugin. Namely `cache` and `cacheOffsetInSecs` (their role is described above)

## Security and Privacy

The auth token received from IAM is now stored in a cache on the user side. This creates more responsibility for the library client to ensure the token is not compromised or leaked.

## Testing

* Added new test `supports an IAM token cache` in `test/plugins/iamauth.js`

## Monitoring and Logging

* Added new log line on line 42 of `lib/tokens/IAMTokenManager.js` to log when we are using a cached token.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# 4.3.0 (2020-07-16)
+- [NEW] Added support for a cache to be optionally used in conjunction with the
+  iamauth plugin, in order to store and retrieve an IAM auth token and lower 
+  the number of IAM token exchanges.
+- [UPGRADED] Upgraded lodash to 4.17.19
+
 # 4.2.4 (2020-03-02)
 - [FIXED] Pinned Nano to version 8.1 to resolve issue with extending upstream
   TypeScript changes in Nano version 8.2.0.
--- a/lib/tokens/IamTokenManager.js
+++ b/lib/tokens/IamTokenManager.js
@@ -59,7 +59,7 @@ class IAMTokenManager extends TokenManager {
             'apikey': self._iamApiKey
           },
           json: true
-        }, (error, response, body) => {
+        }, async(error, response, body) => {
           if (error) {
             callback(error);
           } else if (response.statusCode === 200) {
--- a/lib/tokens/IamTokenManager.js
+++ b/lib/tokens/IamTokenManager.js
@@ -17,6 +17,8 @@ const a = require('async');
 const debug = require('debug')('cloudant:tokens:iamtokenmanager');
 const TokenManager = require('./TokenManager');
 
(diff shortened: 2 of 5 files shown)


hermansb:
@dscape @smithsz @ricellis would one of you be able to take a look at this (or poke someone that would be able to)?

ricellis:
>issue of excessive IAM API key exchanges for access tokens

Do you know what version was running when you saw the issue? It looks like the [loopback connector is set to 4.2.1 minimum](https://github.com/strongloop/loopback-connector-cloudant/blob/871338965737234e2e80b0a97b7879c91f2112fe/package.json#L33).

Prior to 4.2.2 the client would wait for a token to expire before renewing and then every request would attempt to renew the token simultaneously, which could create a lot of requests at once. That was fixed in 4.2.2 so that there was only 1 request per client instance and the renewal requests are made pre-emptively (at 50% of token lifetime) - so the expectation is one token exchange per client instance every half-hour.

If you still really want to cache tokens between client instances my recommendation would be to consider doing so via setting the `iamTokenUrl` to point to a proxy where you run a cache. This would have the advantage of making the cache shareable more widely than this approach and usable by other IAM based resources (e.g. not limited to nodejs and Cloudant). Since most other IAM helpers already accept a custom token auth URL this will also make the cache more future-proof. For example, this library will be superseded by an [IBM Cloud SDK](https://github.com/IBM/ibm-cloud-sdk-common/) where these plugin modifications would not be usable, but a cache server could still be used.

ricellis:
For the record another alternative is to keep a cache independently updated with fresh tokens and use a simpler custom plugin to fetch the token only from the cache and apply it to each request in a `Authorization: Bearer <token>` header.

PR #440, opened 2021-04-01
Change timeout handling in TokenManager

vmatyusGitHub:
## Checklist

- [ ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [ ] Added tests for code changes _or_ test/build only changes
- [ ] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [ ] Completed the PR template below:

## Description

According to the snapshot created by Eric, clearly can be seen there is a `setTimeout` that binds data to the native context.

> Native objects are everything else which is not in the JavaScript heap.

Screenshot 2021-04-01 at 15 06 45

This code part refers to this:
```javascript
setTimeout(this._autoRenew.bind(this, defaultMaxAgeSecs), delayMSecs).unref()
```

Here is small code example that produce the same memory trace:
```javascript
var heapdump = require('heapdump');

function sleep(seconds){
    var waitUntil = new Date().getTime() + seconds*1000;
    while(new Date().getTime() < waitUntil) true;
}

function print() {
    console.log("Hello")
}

setTimeout(print, 10).unref();

sleep(2)

heapdump.writeSnapshot('./timeout-unref.heapsnapshot');
```

This produce the same memory trace:
Screenshot 2021-04-01 at 15 35 15

## Approach

Extending the sample code with a timer clear looks like this:

```javascript
var heapdump = require('heapdump');

function sleep(seconds){
    var waitUntil = new Date().getTime() + seconds*1000;
    while(new Date().getTime() < waitUntil) true;
}

function print() {
    console.log("Hello")
}

myTimer = setTimeout(print, 10);

sleep(1)

clearTimeout(myTimer)
myTimer = setTimeout(print, 10);

sleep(1)

clearTimeout(myTimer)

heapdump.writeSnapshot('./timeout-var.heapsnapshot');
```

And this code memory trace looks this way:
Screenshot 2021-04-01 at 15 52 43

This means that the referenced object will be placed into the `global` not into the `system/context` scope.

Hopefully this change will allow for the native objects to get cleared by GC.

## Testing

Travis has some pending problem, but do not know why.

### DIFF
--- a/lib/tokens/TokenManager.js
+++ b/lib/tokens/TokenManager.js
@@ -27,18 +27,20 @@ class TokenManager {
     this._isTokenRenewing = false;
 
     this._tokenExchangeEE = new EventEmitter().setMaxListeners(Infinity);
+    this._renewTimeout = {};
   }
 
   _autoRenew(defaultMaxAgeSecs) {
     debug('Auto renewing token now...');
+    clearTimeout(this._renewTimeout);
     this._renew().then((response) => {
       let maxAgeSecs = cookie.parse(response.headers['set-cookie'][0])['Max-Age'] || defaultMaxAgeSecs;
       let delayMSecs = maxAgeSecs / 2 * 1000;
       debug(`Renewing token in ${delayMSecs} milliseconds.`);
-      setTimeout(this._autoRenew.bind(this, defaultMaxAgeSecs), delayMSecs).unref();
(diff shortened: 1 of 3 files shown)


vmatyusGitHub:
This correction was not enough to remove the memory leak.

PR #455, opened 2021-06-09
454 double callback

### LINKED ISSUE #454: Unhandled promise rejection with `Error: Callback was already called.`
## Bug Description

Customer reported an unhandled promise rejection:
```
Error: Callback was already called.
    at /opt/app-root/src/node_modules/async/dist/async.js:985:32
    at updateState (/opt/app-root/src/node_modules/@cloudant/cloudant/lib/clientutils.js:53:3)
    at /opt/app-root/src/node_modules/@cloudant/cloudant/lib/clientutils.js:142:11
    at /opt/app-root/src/node_modules/@cloudant/cloudant/plugins/iamauth.js:88:7
```

The problem was reported from a production app without debug enabled. Only the `iamauth` plugin is active.

One callback at the completion of the plugin async eachseries.

### 3. What actually happened

Two callbacks.
Based on the stack my best guess is that it is possible for an error during a success callback to cause an error that will in turn be caught by the promise chain and trigger a second callback.

## Environment details

- Version(s) that are affected by this issue.
    > 4.3.0
- Node.js version
    > 14.16.0

ricellis:
## Checklist

## Description

Correct opportunities for double callbacks

Fixes #454 

## Approach

When writing tests I noticed that max-age was not correctly parsed from a single set cookie header (only from an array) - so there is one fix to resolve that issue.

Double callback (case 1):
In auth plugins for case where an error is encountered during state update.
The `cookieauth` and `iamauth` plugins are modified to `callback(state)` in a `finally` at the end of a promise chain instead of from two places in the `onRequest` function.

Double callback (case 2):
Discovered when updating nock to 13.x (see Test section below). In the case a request is aborted for a retry a double request `end` event might be received because (to [more precisely mirror current Node](https://github.com/nock/nock/blob/main/migration_guides/migrating_to_13.md#upgrading-from-nock-12-to-nock-13) behaviours the `abort` event moved into the next tick. To ensure our plugins behave correctly with the aborts the callbacks in `lib/clientutils.js` `processState` are deferred to the next tick.

## Schema & API Changes

- Remove EOL Node.js versions from test matrix and supported engines

## Security and Privacy

- Fix package-lock.json for a npm audit warning

## Testing

* Moved `test/mocha.opts` -> `.mocharc.json` to avoid warning for using deprecated format.
* Existing tests provide good coverage that functionality is unaffected.
    * For the case 1 double callback I could monkey-patch the state update to throw an error to demonstrate the problem (and did so during manual testing), but I don't think it would provide any value in automated test.
* As I was initially investigating I wrote a couple of extra tests for some IAM auth error cases.
* Upgraded nock to 13.x (to resolve deprecation warnings from Node versions >=10)
     * Caused `CloudantClient/#db using listeners/with ComplexPlugin2` to fail with the `end` event being triggered twice, causing two done callbacks. This is similar to the reported issue so another fix (described above) was introduced to resolve the failure.
     * This test covers the case 2 double callback

## Monitoring and Logging

- "No change"

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,5 @@
+{
+"exclude": "test/typescript/*.js",
+"recursive": true,
+"timeout": 60000
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ branches:
     - rewrite
 node_js:
   - "node"
-  - "10"
-  - "8"
+  - "14"
+  - "12"
 os:
   - linux
 before_install:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Unreleased
 - [FIXED] Parsing of max-age from Set-Cookie headers.
 - [FIXED] Double callback if plugin errors when updating state.
+- [NOTE] Updated engines to remove EOL Node.js versions (minimum is now Node.js 12).
 
 # 4.3.1 (2021-03-17)
 - [NEW] Add migration guide to the newly supported cloudant-node-sdk
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # Unreleased
 - [FIXED] Parsing of max-age from Set-Cookie headers.
+- [FIXED] Double callback if plugin errors when updating state.
 
 # 4.3.1 (2021-03-17)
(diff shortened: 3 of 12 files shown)


vmatyusGitHub (approved):

emlaver (approved):
Looks good!

vmatyusGitHub (approved):

PR #457, opened 2021-07-07
456 retry timeout hang

### LINKED ISSUE #456: nodejs-cloudant hangs when response from cloudant is slow
We use nodejs-cloudant 4.2.2. And this can be reproduced on 4.4.0.
Cloudant met DoS attack some days ago, and we found nodejs-cloudant hangs during this time in our production env.
We also reproduced this by using following https server instead of cloudant server.
```
var app = https.createServer(options, function(req, res){
  console.log("=====================Helllo world.======================");
  let body = 'hello world'
  res.write(body);
//  res.end();
}).listen(443, "0.0.0.0");
```
In reproduce situation, our timeout is 30000ms. In production env, we do not set timeout.
I debug the nodejs-cloudant in reproduce case with timeout. And found in https://github.com/cloudant/nodejs-cloudant/blob/263dddaa07f1c50d8cc1fe341646cfbea574d87c/lib/client.js#L230:
```
request.response
      .on('response', function(response) {
        request.response.pause();
        utils.runHooks('onResponse', request, response, function() {
          utils.processState(request, done); // process response hook results
        });
      });
```
The parameter of done callback in utils.processState is "new Error('No retry requested')" which causes async.forever exits.
When timeout abort emit, nodejs-cloudant run the code section from the line "https://github.com/cloudant/nodejs-cloudant/blob/263dddaa07f1c50d8cc1fe341646cfbea574d87c/lib/clientutils.js#L158" to try to retry the request. Since async.forever has exit, nodejs-cloudant hangs there.

It seems "https://github.com/cloudant/nodejs-cloudant/blob/263dddaa07f1c50d8cc1fe341646cfbea574d87c/lib/clientutils.js#L167" and "https://github.com/cloudant/nodejs-cloudant/blob/263dddaa07f1c50d8cc1fe341646cfbea574d87c/lib/clientutils.js#L163" should be the best place to exit async.forever. Exit aync.forever in reponse hooks may cause problem when response is emit by request package but network problem happend.

Because my reproduce is not based on the real cloudant send response scenario, and i am not very familiar with request package, my analysis may be not right. 
But please check if there is some problem in nodejs-cloudant even if we add timeout in our code explicitly.  We had met such hang issues in our production env 3 times in last and this month. And this also cause our service can not restore even if cloudant server restores. We must restart our service manually to make it work.

ricellis:
## Checklist

## Description

Ensure callback gets called for errors after response is already being processed.

Fixes #456 

## Approach

* Add a flag when the execution loop is stopped.
* Force the client callback when processing a callback `error` if the flag is set.
* Update documentation to clarify behaviour of retry plugin.

It is worth noting the current behaviour on an error encountered during response processing when using the retry plugin is:
* The retry plugin asks for a retry, but it does not get executed.
* Callback style -> hangs with the callback not called
* Listener style -> an event is emitted for the error (between the `response` and `end` events)

This change retains the current behaviour except for the hang where the callback isn't called. It is not possible to allow retries once the response is already processing because it will change the events in the listener style usage (and multiple response or data events will be unexpected and difficult to manage).

## Schema & API Changes

- No change
- Added documentation clarification for expected behaviour of retry plugin.

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - `does not retry in response data phase` for `callback only` `listener only` and `callback and listener` scenarios of retry plugin tests

The tests cannot use `nock` because it does not trigger the timeouts, so they use their own mock server.

## Monitoring and Logging

- Added debug statements to retry plugin.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 # UNRELEASED
+- [FIXED] Hang caused by plugins (i.e. retry plugin) preventing callback execution
+  by attempting to retry on errors received after starting to return the response body.
 - [DEPRECATED] This library is now deprecated and will be EOL on Dec 31 2021.
 
 # 4.4.0 (2021-06-18)
--- a/README.md
+++ b/README.md
@@ -511,7 +511,10 @@ var cloudant = Cloudant({ url: myurl, maxAttempt: 5, plugins: [ { iamauth: { iam
     - `retryErrors`
 
       Automatically retry a request on error (e.g. connection reset errors)
-      _(default: true)_.
+      _(default: true)_. Note that this will only retry errors encountered
+      _before_ the library starts to read response body data. After that point
+      any errors (e.g. socket timeout reading from the server) will be returned
+      to the caller (via callback or emitted `error` depending on the usage). 
 
     - `retryInitialDelayMsecs`
 
--- a/lib/clientutils.js
+++ b/lib/clientutils.js
@@ -120,6 +120,7 @@ var processState = function(r, callback) {
   }
 
   // [5] => Return response to awaiting client.
+  r.state.final = true; // flags that we are terminating the loop
   setImmediate(callback, new Error('No retry requested')); // no retry
 };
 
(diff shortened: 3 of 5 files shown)


mojito317 (approved):
LGTM!

emlaver (approved):

